PR #4, opened 2022-01-16
Reduced search bar motion. Search button focuses cursor.

rosettet:
- Search button no longer shifts to the right when clicked
- Navigation bar is slightly taller, no longer changing height when the search bar is hidden/revealed.
- Clicking the search button now focuses the cursor in the text box, saving precious milliseconds.


Absolucy:
Thanks for your contribution! <3

PR #5, opened 2022-01-19
WiFi scanning improvements

mmstick:
Something like this should work effectively for handling GTK events in Rust without the need for much reference counting.

- Uses an event loop in a `glib::MainContext::default()` spawned local future for handling events to the WiFi widget and managing all WiFi-related state.
- Sends events using a `tokio::sync::mpsc::UnboundedSender` so it doesn't block widget signals
- AP info stored in a `SlotMap` so they can be referenced by an ID across events.
- Added dummy dialog to respond to configure signals
- Convenience function `task::spawn_local` for spawning tasks locally on GLib.
- `task::spawn` for spawning tasks in the background on Tokio.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -304,6 +304,7 @@ dependencies = [
  "once_cell",
  "os-release",
  "relm4-macros",
+ "slotmap",
  "sysinfo",
  "tokio",
  "tokio-stream",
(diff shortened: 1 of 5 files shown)


Absolucy (approved):
Thanks!

wash2 (commented):
Thanks, this approach looks good

PR #22, opened 2023-01-31
translation to Brazilian Portuguese

marccdb:
Happy to help! =)


leviport:
It looks like there was a contribution collision, https://github.com/pop-os/cosmic-settings/pull/21 looks like it does the same thing

marccdb:
> It looks like there was a contribution collision, #21 looks like it does the same thing

Indeed, just a few minutes apart.

PR #26, opened 2023-02-17
add italian language translation

Krahos:
Notes for the translation:
- I translated "dock" as "barra applicazioni" (which litterally translates to application bar): in my opinion it's more understandable than leaving the word "dock" like MacOS does, which is the only real alternative to this translation.
- I didn't translate "hot corners": same name is used on Windows for Italian. The litteral translation would be "angoli caldi", but it doesn't sound too great.
- I chose the word "avviatore" for "launcher": "lanciatore", the litteral translation, is not very used in Italian. "Avviatore" means "something that starts something else" and sounds better in the context. Another translation could be "starter": it's an English word, but it's a very understandable word. The alternative is to call it "launcher" in Italian as well.
- I chose to translate "memory" to "RAM" and "disk capacity" to "memoria di massa". I think it's more clear for a user what they refer to and I personally would change them in English as well, especially since mass storage is rarely disks nowdays.


jacobgkau (approved):
Built locally to test since the build server can't build this package right now (also affecting the master branch, so not a regression.) The Italian translation is showing up properly in a user account set to Italian.

Thank you for the contribution! I'm merging this now, but it would be greatly appreciated if you could visit this again later in development, as there will likely be many new and changed strings.

PR #27, opened 2023-02-24
i18n(sr): Add Serbian (Latin) translation

AAnzel:


jacobgkau (approved):
Thank you for the contribution! This is showing up successfully in a user account set to Serbian (Latin).

PR #30, opened 2023-04-29
settings deutsch (de)

soloturn:


jacobgkau (approved):
Thank you for the contribution! This is showing up successfully in a user account set to German.

jacobgkau:
Force-pushed to make the commit message match other translation commit messages.

PR #31, opened 2023-04-29
Add Hindi Translation

178900PRITAM:


jacobgkau (approved):
Thank you for the contribution! This is showing up successfully in a user account set to Hindi.

jacobgkau (approved):

jacobgkau:
Force-pushed to make the commit message match other translation commit messages.

PR #39, opened 2023-05-25
Cosmic panel jammy

wash2:
This adds a lot of panel configuration to cosmic-settings and uses iced-sctk. A couuple items in the panel page are TODO but I figured a PR for the switch to iced-sctk is better sooner than later.

Also, I think a couple improvements could be added around handling of window states for un-rounding the corners of the window headerbar when maximized ,fullscreened, or tiled. Iced-sctk needs to forward window state events for that to happen, but that's a small change.


mmstick:
Is this good to merge?

wash2:
Yes, I think so

PR #44, opened 2023-06-07
Fix build jammy

wash2:
seems to be succeeding now


mmstick (commented):
Is there an issue with 1.70?

wash2:
Is there an issue with 1.70?

I think only 1.69.0 is available to be installed for rust-all right now as far as I can tell

PR #46, opened 2023-06-12
feat: theme management

wash2:
This loads and watches the theme for changes


mmstick (approved):
I wonder if there's a simpler one-liner we could use so app development is easier

wash2:
Ya, I made a helper method for applets and I could probably make one for regular apps and include it in libcosmic. It wouldn't be one line, but it could be a lot less 😅

PR #49, opened 2023-06-15
udpate libcosmic

wash2:
seems to work well when i test it :)


mmstick (approved):
Not seeing any issues here either

PR #58, opened 2023-06-26
feat: add nix flake

wash2:


flake.nix
```diff
@@ -0,0 +1,86 @@
...
+          };
+          nativeBuildInputs = with pkgs; [
+            just
+            pkg-config
+            autoPatchelfHook
+          ];
+          buildInputs = with pkgs; [
+            libxkbcommon
+            glib
+            gtk4
```
mmstick, line 48: Is glib and gtk4 necessary? We only need these for debian packaging:

- cmake
- libexpat1-dev
- libfontconfig-dev
- libfreetype-dev lld
- pkg-config
- libxkbcommon-dev

wash2: Ahh ya, I mostly just copied one of our exisitng flakes for this. I'll see if i can clean it up a bit.

mmstick (approved):

PR #60, opened 2023-06-27
Wallpaper settings fixes & improvements

mmstick:
- Fixes the stack overflow crash when loading the config
- Reduces memory usage by 90% when loading backgrounds for the first time
- Simulates how the selected wallpaper will look like for the given scaling mode.


wash2:
Something weird that i've just noticed about master_jammy and this branch is that the scrollbar on the wallpaper page is tiny because the height of the page is something huge.

mmstick:
Yes, the existence of a responsive widget causes this issue.

wash2 (approved):

PR #61, opened 2023-06-28
Add input settings, with mouse, keyboard, keyboard shortcuts sub-pages

ids1024:
Still needs to be integrated with compositor so it actually works, need touchpad page, etc.


ids1024:
With the current state of this, and the changes now merged in cosmic-comp, this now has working touchpad and mouse settings. Keyboard settings lets you set a compose key, but the UI for selecting from possible keyboard layouts is still needed. Keyboard shortcuts are still needed.

This is good start anyway.

mmstick (approved):

PR #62, opened 2023-06-28
WIP update workspace setttings to match mockup

ids1024:
Still need to start implementation functionality. Radio buttions don't easily work quite as the mockup shows.


ids1024:
The styling of everything still isn't quite like the mockup, and the images need to be added, but it seems worth merging since the settings are useful. It can be improved in future PRs.

wash2 (approved):

PR #64, opened 2023-07-18
i18n(tr): add Turkish language

ysfchn:
Added Turkish language.

"Applets" are translated as "Extensions", because there is no exact translation of "applet" in Turkish, the closest word could be "uygulamacık" (means "tiny-apps", "mini-apps"), but "uygulamacık" sounded too silly to me and I thought it could be confused with the actual apps that you install on the system.

I'm also hoping to translate the new strings once they added!

### DIFF
--- a/i18n/tr/cosmic_settings.ftl
+++ b/i18n/tr/cosmic_settings.ftl
@@ -1,0 +1,223 @@
+app = COSMIC Ayarları
+
+unknown = Bilinmiyor
+
+number = { $number }
+
+## Desktop
+
+desktop = Masaüstü
+
+## Desktop: Appearance
+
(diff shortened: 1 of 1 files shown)


mmstick:
We should find a different term for applets because they aren't extensions. COSMIC applets are small applications which integrate with the compositor as a shell component. So "Shell Components" might be a suitable replacement.

mmstick (approved):

ysfchn:
I'm glad I could contribute to the project!

I hope "kabuk bileşenleri" ("shell components") don't cause a visual pollution and can fit to the UI due to its long-a-bit name. :sweat_smile: But the translation can be of course improved by the time.

PR #76, opened 2023-09-19
refactor: libcosmic rebase with Application API

mmstick:
Requires https://github.com/pop-os/libcosmic/pull/165

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -856,7 +856,7 @@ dependencies = [
  "futures-lite",
  "image",
  "rayon",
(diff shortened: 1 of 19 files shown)


wash2:
I'm not yet sure about the cause, but reordering applets with DnD seems to not quite finish properly when I test this branch.

wash2:
I pushed a small fix that should work for now, but I think something is off about the DnD events and requests. I'll look into it more when I have some time.

wash2:
https://github.com/pop-os/iced/pull/60 should probably help fix the build

wash2 (approved):

PR #78, opened 2023-09-22
Appearance jammy

wash2:
WIP appearance page

### DIFF
--- a/app/src/app.rs
+++ b/app/src/app.rs
@@ -19,7 +19,7 @@ use cosmic_settings_page::{self as page, section};
 use crate::config::Config;
 
 use crate::pages::desktop::{
-    self,
+    self, appearance,
     dock::{self, applets::ADD_DOCK_APPLET_DIALOGUE_ID},
     panel::{
         self,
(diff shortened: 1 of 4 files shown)


jackpot51:
I think there is some multi-window issue. After opening the color picker and picking a color, the picker closes and then the whole app panics on line 1363 in iced/sctk/src/application.rs

wash2:
> I think there is some multi-window issue. After opening the color picker and picking a color, the picker closes and then the whole app panics on line 1363 in iced/sctk/src/application.rs

I've not seen this happening yet after using it. Is there a specific color picker or method of closing the color picker window that causes the issue?

jackpot51:
I'm not having any panics with the latest update. I'm seeing the custom items below accent color all show up as black, when set to the defaults:

Screenshot_2023-10-11_12-26-44

wash2:
> I'm not having any panics with the latest update. I'm seeing the custom items below accent color all show up as black, when set to the defaults:
> 
> Screenshot_2023-10-11_12-26-44

Ya, my understanding is that if nothing is selected, the color picker should appear black. Maybe it should be transparent, or I can try setting the color to the default of the theme if it is left out.

wash2:
Oh actually, maybe it could be good if the color pickers in general display the + when no color is selected, like the accent color picker in figma?

mmstick (approved):

PR #87, opened 2023-11-04
Update Russian translation

FAlexei:
Russian translation update with new strings


jackpot51 (approved):
Thanks for the update!

PR #92, opened 2023-11-13
refactor(appearance): use image button in the appearance panel

wash2:


mmstick:
PR is closed, so it'll need to be opened to review

PR #93, opened 2023-11-13
refactor(appearance): use image button in the appearance panel

wash2:


mmstick (approved):
We should eventually migrate the color image button widget to libcosmic

PR #128, opened 2024-01-29
Update libcosmic with some wallpaper page improvements

mmstick:
- Updates libcosmic, which had changes to file chooser dialogs and config subscriptions
- Remove stretch feature from the wallpaper settings UI
- Renames Zoom to Fill in wallpaper settings UI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -309,9 +309,9 @@ dependencies = [
 
 [[package]]
 name = "async-io"
-version = "2.3.0"
+version = "2.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fb41eb19024a91746eba0773aa5e16036045bbf45733766661099e182ea6a744"
+checksum = "8f97ab0c5b00a7cdbe5a371b9a782ee7be1316095885c8a4ea1daf490eb0ef65"
 dependencies = [
  "async-lock 3.3.0",
  "cfg-if",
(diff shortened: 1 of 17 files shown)


wash2:
I think the last commit to the panel should fix the sctk conflicts when vendoring, and undoing the update to the commit should fix the build.

gabriele2000:
> * Remove stretch feature from the wallpaper settings UI
> 
> * Renames Zoom to Fill in wallpaper settings UI

Since you're removing the stretch feature...

"Zoom" has this issue while "Stretch" is perfectly fine.
screenshot-2024-01-29-19-14-22

mmstick:
@gabriele2000 Not able to replicate. Can you shall the image file?

gabriele2000:
> @gabriele2000 Not able to replicate. Can you share the image file?

Wallpaper

mmstick:
@gabriele2000 I was able to replicate with that specific image, though it doesn't affect any of the other images I've tested with. It seems to be a bug within the version of the image crate used by the `cosmic-bg` daemon, but I already planned to update the daemon so this might be resolved once I finish updating it.

mmstick:
@gabriele2000 I was able to resolve the issue in https://github.com/pop-os/cosmic-bg/pull/27

I don't know the cause of the issue though. I noticed that when calling `imageops` functions directly, it scales without issue. But if using resize method from `DynamicImage`, it's broken. This might be a possible test case for reporting a bug in the image crate.

PR #156, opened 2024-02-17
add some french translation

wiiznokes:


wiiznokes:
hum, do you know a tool i could use to see what fields are needed or need to be removed ?

mmstick:
Not that I know of. The recent changes are to the input devices pages.

mmstick (approved):

wiiznokes:
I'm not sure about the translation of panel in frensh. I translated it into "Barre d'Applets", but it is opinionated, after thinking, maybe "panneau" is preferable, but in my head, i see a "sign" when i hear "panneau". I'm not sure, but this could be changed after anyway

Issue #164, opened 2024-02-26
Settings > About: Double check calculations; Truncate calculated values.


XV-02:
Looking at the values for `Disk Capacity` and `Memory`, I noticed that we're calculating out eight or more decimal places when converting from metric gig to true gig.

![screenshot-2024-02-26-18-10-22](https://github.com/pop-os/cosmic-settings/assets/98765732/45adf7cb-8c59-4e65-b3d5-8d4644c1e7d3)

I think we should probably truncate to 2 decimal places. This will have the effect of rounding all values down.

We could do this by truncating to 2 non-zero places:

| Full Value | Truncated Value |
| -------------- | ------------------------ |
| 256.0000  | 256                       |
| 249.0017  | 249                       |
| 253.4633  | 253.46                  |
| 202.6076  | 202.6                    |

Alternatively, just truncating to two decimal places full stop:

| Full Value | Truncated Value |
| -------------- | ------------------------ |
| 256.0000  | 256.00                  |
| 249.0017  | 249.00                  |
| 253.4633  | 253.46                  |
| 202.6076  | 202.60                  |

In either case, we avoid over-stating capacity, and we make the information more visually immediate.

We also need to make sure that the math we're doing for the `Memory` value is correct, as the Pang12 has 32 metric gig of memory, but is reporting _over_ that value, which it shouldn't. If I look at the value elsewhere - such as `htop` - I see 30.6G or 30.5G reported depending on rounding. The conversion should always indicate a lower true value than marketing value.

PR #195, opened 2024-03-12
system/about: Improve memory and disk capacity display

coolavery:
This PR fixes the two issues brought up in #164 with how memory and disk capacity are displayed:

- The values are now displayed in binary units (GiB, TiB, etc.) instead of decimal/metric units (GB, TB, etc.)
- The values are now rounded to two decimal places

| Before | After |
| --- | --- |
| Memory: 33.555890176 GB, Disk Capacity: 13.800913174528 TB | Memory: 31.25 GiB, Disk Capacity: 12.55 TiB |

Previously this page was displaying metric/decimal units, which is why the capacity seemed overstated.

I also rounded to two decimals places, instead of truncating, as recommended by the original issue. This is mostly because it requires less code, although it seems like the original issue suggested truncating to prevent showing more than the advertised memory/disk space. This should no longer be an issue now that binary units are displayed, and all the other tools I checked (htop, KDE System Monitor, Mission Centre) show a rounded up value.

### DIFF
--- a/pages/system/src/about.rs
+++ b/pages/system/src/about.rs
@@ -209,7 +209,8 @@ pub fn read_to_string<'a, P: AsRef<OsStr>>(
 }
 
 fn format_size(size: u64) -> String {
-    byte_unit::Byte::from_u64(size)
-        .get_appropriate_unit(byte_unit::UnitType::Binary)
-        .to_string()
+    format!(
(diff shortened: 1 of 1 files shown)


mmstick (approved):

mmstick:
Thanks!

PR #220, opened 2024-03-27
Keyboard settings page

mmstick:
- [x] Adds inner-app dialog API support for settings pages
- [x] Migrate keyboard layout code from distinst
- [x] Move keyboard settings data to keyboard page
- [x] Convert keyboard dialogs into contexts for the context drawer
- [x] Refactor to reference layouts by IDs
- [x] Add context drawer to add new input sources
- [x] Searchable input sources
- [x] Implement context popover buttons

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1328,6 +1328,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "url",
+ "xkb-data",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


leb-kuchen:
Is it not insecure, if setting a keyboard shortcut doesn't require root, since you could create infinite shortcuts to key log you?

ids1024:
> Is it not insecure, if setting a keyboard shortcut doesn't require root, since you could create infinite shortcuts to key log you?

Standard Unix permissions generally do not protect you against a keylogger running under your account. For instance https://github.com/Aishou/wayland-keylogger. (Despite the name, this isn't an issue with *Wayland*. `LD_PRELOAD` could inject key logging into other things).

(Though keyboard shortcuts wouldn't really be a way to implement a key logger. Generally you can't bind a keyboard shortcut for just a key without modifiers. Shortcuts also intercept the key combination so it isn't sent to apps.)

If you want to protect against things like that, you need sandboxing. Like Flatpak (with appropriate restrictions on permissions).

Issue #208, opened 2024-03-20
Add icon themes previews to menu screen


TheESN:
![main](https://github.com/pop-os/cosmic-settings/assets/102585253/846083af-fa1e-47d8-9cb9-3156b6b73b4e)

I suggest to make something like in KDE Plasma to make it more clear to user what exactly icons they want to choose

joshuamegnauth54:
I implemented a feature in the linked PR that seems _sort of_ like this.

maria-komarova:
#235 should describe the design for the feature that should address this issue once implemented.

PR #232, opened 2024-04-09
feat: Preview icon themes

joshuamegnauth54:
(Maybe) closes: #208 

icon_picker

Edit: Actually it may be a little different than #208 but it still seems pretty useful.

### DIFF
--- a/cosmic-settings/src/pages/desktop/appearance.rs
+++ b/cosmic-settings/src/pages/desktop/appearance.rs
@@ -36,7 +36,9 @@ use super::wallpaper::widgets::color_image;
 
 const ICON_PREV_N: usize = 6;
 const ICON_PREV_ROW: usize = 3;
-const ICON_PREV_SIZE: u16 = 32;
+const ICON_TRY_SIZES: [u16; 3] = [32, 48, 64];
+const ICON_THUMB_SIZE: u16 = 32;
+const ICON_NAME_TRUNC: usize = 20;
 type IconThemes = Vec<String>;
 type IconHandles = Vec<[icon::Handle; ICON_PREV_N]>;
 
@@ -1655,12 +1657,18 @@ fn preview_handles(theme: String) -> [icon::Handle; ICON_PREV_N] {
 }
 
 fn icon_handle(icon_name: &str) -> icon::Handle {
-    icon::from_name(icon_name)
-        // Get the path to the icon for the currently set theme.
-        // Without the exact path, the handles will all resolve to icons from the same theme in
-        // [`icon_theme_button`] rather than the icons for each different theme
-        .path()
-        .map(icon::from_path)
+    ICON_TRY_SIZES
+        .iter()
+        .find_map(|&size| {
+            icon::from_name(icon_name)
+                // Set the size on the handle to evaluate the correct icon
+                .size(size)
+                // Get the path to the icon for the currently set theme.
+                // Without the exact path, the handles will all resolve to icons from the same theme in
+                // [`icon_theme_button`] rather than the icons for each different theme
+                .path()
+                .map(icon::from_path)
+        })
         // Fallback icon handle
         .unwrap_or_else(|| icon::from_name(icon_name).handle())
 }
@@ -1687,7 +1695,7 @@ fn icon_theme_button(
                                 .take(ICON_PREV_ROW)
                                 .cloned()
                                 // TODO: Maybe allow choosable sizes/zooming
-                                .map(|handle| handle.icon().size(ICON_PREV_SIZE)),
+                                .map(|handle| handle.icon().size(ICON_THUMB_SIZE)),
                         )
                         .spacing(theme.space_xxs())
                         .into(),
@@ -1698,7 +1706,7 @@ fn icon_theme_button(
                                 .skip(ICON_PREV_ROW)
                                 .cloned()
                                 // TODO: Maybe allow choosable sizes/zooming
-                                .map(|handle| handle.icon().size(ICON_PREV_SIZE)),
+                                .map(|handle| handle.icon().size(ICON_THUMB_SIZE)),
                         )
                         .spacing(theme.space_xxs())
                         .into(),
@@ -1752,12 +1760,12 @@ fn icon_theme_button(
             }),
         )
         .push(
-            text(if name.len() > 18 {
-                format!("{name:.20}...")
+            text(if name.len() > ICON_NAME_TRUNC {
+                format!("{name:.ICON_NAME_TRUNC$}...")
             } else {
                 name.into()
             })
-            .width(Length::Fixed((ICON_PREV_SIZE * 3) as _)),
+            .width(Length::Fixed((ICON_THUMB_SIZE * 3) as _)),
         )
         .spacing(theme.space_xs())
         .into()
--- a/cosmic-settings/src/pages/desktop/appearance.rs
+++ b/cosmic-settings/src/pages/desktop/appearance.rs
@@ -3,7 +3,7 @@
 
 use std::borrow::Cow;
 use std::collections::BTreeMap;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::sync::Arc;
 
 use ashpd::desktop::file_chooser::{FileFilter, SelectedFiles};
@@ -13,13 +13,13 @@ use cosmic::cosmic_theme::palette::{FromColor, Hsv, Srgb, Srgba};
 use cosmic::cosmic_theme::{
     CornerRadii, Theme, ThemeBuilder, ThemeMode, DARK_THEME_BUILDER_ID, LIGHT_THEME_BUILDER_ID,
 };
-use cosmic::iced_core::{alignment, Color, Length};
-use cosmic::iced_widget::{scrollable, Column};
+use cosmic::iced_core::{alignment, Background, Color, Length};
+use cosmic::iced_widget::scrollable;
 use cosmic::prelude::CollectionWidget;
 use cosmic::widget::icon::{self, from_name, icon};
 use cosmic::widget::{
-    button, color_picker::ColorPickerUpdate, column, container, horizontal_space, row, settings,
-    spin_button, text, ColorPickerModel,
+    button, color_picker::ColorPickerUpdate, column, container, flex_row, horizontal_space, row,
+    settings, spin_button, text, ColorPickerModel,
 };
 use cosmic::Apply;
 use cosmic::{command, Command, Element};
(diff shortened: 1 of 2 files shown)


mmstick:
Requires design approval. @pop-os/ux

git-f0x:
Personally, it looks a bit busy when always visible like that.
Maybe have icon theme selection in a separate window (similar to the Add Applet window), and in the place of the current drop-down for icon selection, maybe have a clickable name of the current theme, which opens the icon theme selection window.

maria-komarova:
The features have been added before the design was finalized. The design is described here now: [#235](https://github.com/pop-os/cosmic-settings/issues/235). Please let me know if there are any questions.

joshuamegnauth54:
Thank you for your comments, everyone! I'll try to implement these changes.

maria-komarova:
Thanks for working on this!

joshuamegnauth54:
theme_prev01
theme_prev02

I tested it with a few random themes that I placed into a fake `XDG_DATA_DIRS` to make sure it works.

cosmic-settings/src/pages/desktop/appearance.rs
```diff
@@ -439,6 +446,45 @@ impl Page {
             .map(crate::pages::Message::Appearance)
     }
 
+    fn experimental_context_view(&self) -> Element<'_, crate::pages::Message> {
+        let active = self.icon_theme_active;
+        let theme = cosmic::theme::active();
+        let theme = theme.cosmic();
+        cosmic::iced::widget::column![
+            // Export theme choice to GNOME
```
mmstick, line 454: This option will apply to other toolkits as well. Not just GTK/GNOME.

cosmic-settings/src/pages/desktop/appearance.rs
```diff
@@ -1571,3 +1631,102 @@ async fn set_gnome_icon_theme(theme: String) {
...
+    handles: &[icon::Handle],
+    id: usize,
+    selected: bool,
+) -> Element<'static, Message> {
+    let theme = cosmic::theme::active();
+    let theme = theme.cosmic();
+    cosmic::iced::widget::column![
+        button(
+            cosmic::iced::widget::column![
+                row::Row::new()
```
mmstick, line 1679: use `cosmic::widget::row()` instead.

cosmic-settings/src/pages/desktop/appearance.rs
```diff
@@ -1571,3 +1631,102 @@ async fn set_gnome_icon_theme(theme: String) {
...
+/// Button with a preview of the icon theme.
+fn icon_theme_button(
+    name: &str,
+    handles: &[icon::Handle],
+    id: usize,
+    selected: bool,
+) -> Element<'static, Message> {
+    let theme = cosmic::theme::active();
+    let theme = theme.cosmic();
+    cosmic::iced::widget::column![
```
mmstick, line 1676: Use `cosmic::widget::column()` instead.

cosmic-settings/src/pages/desktop/appearance.rs
```diff
@@ -829,6 +876,13 @@ impl Page {
                 }
                 Command::none()
             }
+            Message::ExperimentalContextDrawer => {
+                self.context_view = Some(ContextView::Experimental);
+
+                cosmic::command::message(crate::app::Message::OpenContextDrawer(
+                    fl!("experimental").into(),
```
mmstick, line 883: From the designs, this should be an empty string

maria-komarova:
Looks great!

Just a few minor things base on the screenshot above (sorry if those were already fixed):
- The copy on the list item on the main content area should say "Experimental settings".
- There should be no header in the Context Drawer or if there is one, then is should say "Experimental Settings".
- The symbolic icon at the end of the list item on the main content area should be "go-next-symbolic" (actions icons).
- "Icon theme" inside Context drawer should be using "Header" text style and it should be spaced from the following thumbnails less (should be @space_xxs). The spacing between "Icon theme" and thumbnails should be less than the spacing between the list item above and "Icon theme" to maintain content hierarchy. 
- We might also want to cut the theme name after two lines. I know there is a text wrapping issue, it's been mentioned in other places. But still better have consistent spacing between rows of thumbnails. We can cut the long names with "...". 

Are icon theme thumbnails using image buttons? Or the same widget wallpapers do? It looks like it has a different corner radius and no additional selection indicator in the left bottom corner.

maria-komarova:
Looks great!

Just a few minor things base on the screenshot above (sorry if those were already fixed):

- The copy on the list item on the main content area should say "Experimental settings".
- There should be no header in the Context Drawer or if there is one, then is should say "Experimental Settings".
- The symbolic icon at the end of the list item on the main content area should be "go-next-symbolic" (actions icons).
- "Icon theme" inside Context drawer should be using "Header" text style and it should be spaced from the following thumbnails less (should be @space_xxs). The spacing between "Icon theme" and thumbnails should be less than the spacing between the list item above and "Icon theme" to maintain content hierarchy.
- We might also want to cut the theme name after two lines. I know there is a text wrapping issue, it's been mentioned in other places. But still better have consistent spacing between rows of thumbnails. We can cut the long names with "...".

Are icon theme thumbnails using image buttons? Or the same widget wallpapers do? It looks like it has a different corner radius and no additional selection indicator in the left bottom corner.

joshuamegnauth54:
menu
context

Here is an update to check if I'm on the right track.

To do tomorrow:
* Header text style
* Shortening long icon titles
* I think I have to pad the buttons a bit

TheESN:
BTW It seems like this screen utilizes the smallest icon size variations from themes. Can you make it use bigger variations?

cosmic-settings/src/pages/desktop/appearance.rs
```diff
@@ -1571,3 +1631,131 @@ async fn set_gnome_icon_theme(theme: String) {
...
+    selected: bool,
+) -> Element<'static, Message> {
+    let theme = cosmic::theme::active();
+    let theme = theme.cosmic();
+    // let image_style = cosmic::theme::Button::Image;
+    let background = theme.palette.neutral_4.into();
+
+    cosmic::widget::column()
+        .push(
+            button::Button::new_image(
```
mmstick, line 1681: This should be `cosmic::widget::button::image()`

cosmic-settings/src/pages/desktop/appearance.rs
```diff
@@ -1571,3 +1631,131 @@ async fn set_gnome_icon_theme(theme: String) {
...
+                cosmic::iced::widget::column![
```
mmstick, line 1682: `cosmic::widget::column()`, `cosmic::widget::column::with_capacity()`, or `cosmic::widget::column::with_children()`.

cosmic-settings/src/pages/desktop/appearance.rs
```diff
@@ -1571,3 +1631,131 @@ async fn set_gnome_icon_theme(theme: String) {
...
+                    cosmic::widget::row::row()
```
mmstick, line 1683: `cosmic::widget::row()`

cosmic-settings/src/pages/desktop/appearance.rs
```diff
@@ -1571,3 +1631,131 @@ async fn set_gnome_icon_theme(theme: String) {
...
+                        .extend(
+                            handles
+                                .iter()
+                                .take(ICON_PREV_ROW)
+                                .cloned()
+                                .map(|handle| handle.icon().size(ICON_PREV_SIZE))
+                        )
+                        .spacing(theme.space_xs()),
+                    row::Row::new()
```
mmstick, line 1692: `cosmic::widget::row()`

cosmic-settings/src/pages/desktop/appearance.rs
```diff
@@ -439,6 +446,48 @@ impl Page {
...
+            // Export theme choice
+            settings::view_section("").add(
+                settings::item::builder(fl!("enable-export"))
+                    .description(fl!("enable-export", "desc"))
+                    .toggler(self.tk.apply_theme_global, Message::ApplyThemeGlobal)
+            ),
+            // Icon theme previews
+            // cosmic::iced::widget::column![text(&*ICON_THEME), text(&*ICON_THEME_DESC).size(10)]
+            //     .spacing(2),
+            cosmic::iced::widget::column![
```
mmstick, line 463: `cosmic::widget::column()`, `cosmic::widget::column::with_capacity()`, or `cosmic::widget::column::with_children()`

mmstick (changes requested):

joshuamegnauth54:
Hello fellow cosmonauts!

Changes for today:
* Truncate long theme names
* Use [heading](https://pop-os.github.io/libcosmic/cosmic/widget/text/fn.heading.html) for "Icon theme"
* I padded the buttons a bit to fit the design better

new01

@TheESN I'm using `32` for the size since it seems like a good default, but I agree that some flexibility will be nice! Maybe a future patch can add a slider to increase or decrease the size?

I can also change the default for now too. I'm aesthetically challenged, so I'm not sure which size is best.

## Icon size 48
size_48

## Icon size 64
size_64

## Icon size 128
:joy_cat: 
size_128

32 (current) and 48 both look pretty nice.

TheESN:
> Hello fellow cosmonauts!
> 
> Changes for today:
> 
>     * Truncate long theme names
> 
>     * Use [heading](https://pop-os.github.io/libcosmic/cosmic/widget/text/fn.heading.html) for "Icon theme"
> 
>     * I padded the buttons a bit to fit the design better
> 
> 
> new01
> 
> @TheESN I'm using `32` for the size since it seems like a good default, but I agree that some flexibility will be nice! Maybe a future patch can add a slider to increase or decrease the size?
> 
> I can also change the default for now too. I'm aesthetically challenged, so I'm not sure which size is best.
> ## Icon size 48
> 
> size_48
> ## Icon size 64
> 
> size_64
> ## Icon size 128
> 
> 😹 size_128
> 
> 32 (current) and 48 both look pretty nice.

I didn't mean the actual size, but their style. Icon themes usually specify several icons for different sizes, so smaller icons have less details, and bigger ones have more details. 

From screenshots that you sent I see that it uses the smallest size of icons and just scales them without changing to more detailed variations.

For reference that's how 32x32 Pop folder should look like
изображение

And that's how 48x48 should look like
изображение

The one that your preview is using is actually 16x16
изображение

joshuamegnauth54:
> I didn't mean the actual size, but their style. Icon themes usually specify several icons for different sizes, so smaller icons have less details, and bigger ones have more details.
> 
> From screenshots that you sent I see that it uses the smallest size of icons and just scales them without changing to more detailed variations.

Alright. I think I fixed it now:

icon_size_fix

It turns out that I was evaluating the sizes too late. I updated the code to check for a few different sizes before falling back to the default too.

The situation for `La Capitaine` and `Luv` is different. I'm unsure if the problem rests with how I manually installed the themes or with `libcosmic`. Basically, the theme name and the directory have to match. `La Capitaine`'s `index.theme` lists the name as `Name=La Capitaine` but the directory is `la-capitaine`. I renamed the directory to match `index.theme` which worked.

The [spec](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html) states:
> Each theme is stored as subdirectories of the base directories. The internal name of the theme is the name of the subdirectory, although the user-visible name as specified by the theme may be different. Hence, theme names are case sensitive, and are limited to ASCII characters. Theme names may also not contain comma or space.

The spec also lists an example of a theme titled `Birch` which is located in a directory called `birch`. I _think_ my issue with `La Capitaine` and `Luv` may be a bug?

TheESN:
Looks definitely better! Though now I am wondering why adwaita is missing settings icon in preview

joshuamegnauth54:
symbolic_fallback

Adwaita doesn't have `preferences-system` but it does have a symbolic iteration of that icon. I updated the code to try the icons listed in the design or fallback to the symbolic variant if the requested icon isn't found.

Alsooo, Vibrancy-Dark was missing icons because I didn't copy over the themes it inherited to my test `XDG_DATA_DIR`. I fixed it to ensure my code still works.

TheESN:
> Adwaita doesn't have preferences-system but it does have a symbolic iteration of that icon

Well now I am wondering how is this icon called then?

изображение

joshuamegnauth54:
I couldn't find an icon that looks like that in `Adwaita`. It's possible that that's an app icon stored in `hicolors` instead of `Adwaita` itself which would allow non-`Adwaita` themes to display the correct icon. `hicolors` is the default fallback for every theme which means it's a useful place to store app icons.

TheESN:
> I couldn't find an icon that looks like that in `Adwaita`. It's possible that that's an app icon stored in `hicolors` instead of `Adwaita` itself which would allow non-`Adwaita` themes to display the correct icon. `hicolors` is the default fallback for every theme which means it's a useful place to store app icons.

Well maybe it's GNOME-only icon then. Kinda weird that they decided to go this way, so symbolic fallback should be fine then

mmstick (approved):

mmstick:
Is this ready for merging?

maria-komarova (approved):

maria-komarova:
Looks ready to me.

Issue #169, opened 2024-02-29
Wallpaper > Slideshow toggle changes to off after being enabled


WatchMkr:
To reproduce
* Activate Slideshow
* Leave the Wallpaper page
* Go back to Wallpaper
* Slideshow is disabled

The wallpapers do appear to change when in slideshow but the toggle is off.

PR #236, opened 2024-04-11
fix: Sync slideshow toggle with cosmic-bg

joshuamegnauth54:
Closes: #169


mmstick (changes requested):
This would be out of sync with cosmic-bg. It already gets slideshow state from cosmic-bg's config. The slideshow feature is enabled when the background is a path.

joshuamegnauth54:
Good point. I updated the code to get the slideshow status from cosmic-bg.

The final code is much simpler than my first attempt.

mmstick (approved):

mmstick:
Thanks!

Issue #211, opened 2024-03-22
Settings > Desktop > Wallpaper: Missing preview and outline of current wallpaper


git-f0x:
When the wallpaper is changed, the preview on the top and the outline with the checkmark on the image are visible:
![screenshot-2024-03-22-15-48-55](https://github.com/pop-os/cosmic-settings/assets/150025636/fc4bc1a6-c327-4517-a8db-744627c5ca15)
If going back from the Wallpaper page and then returning, the outline is missing. Closing Settings and reopening leads to the preview being black, as well.
![screenshot-2024-03-22-15-49-20](https://github.com/pop-os/cosmic-settings/assets/150025636/4ca4db04-b91d-423e-a391-2ed666cda544)

WatchMkr:
This appears to be fixed.

git-f0x:
It's still present for me, unless the latest commit which hasn't yet been pushed to apt fixed it.

edfloreshz:
I cannot reproduce this.

git-f0x:
Why do I always have weird issues... 🫠

WatchMkr:
Just had this happen on an AMD graphics system.

edfloreshz:
Which steps did you take to reproduce it?

git-f0x:
Since I am on AMD graphics, this issue might be related to pop-os/libcosmic#370 , since it also seems to occur on AMD graphics there.

git-f0x:
> Which steps did you take to reproduce it?
For me just opening the Wallpaper page in Settings.

git-f0x:
I seem to have figured out when this happens (at least on my end).
When there is an added image for wallpapers, wallpaper settings don't exhibit this issue.
When there are no added wallpapers, the issue occurs every time.

git-f0x:
This may or may not be related to this: when there's an added wallpaper, it isn't present when the Wallpaper page is first opened. It appears only when something is hovered over, or the window is moved or scrolled.

https://github.com/pop-os/cosmic-settings/assets/150025636/a8fe2a55-b1f7-4bc5-87e7-c9b300abe2fe

gantryyork:
New wallpapers will not be part of the Alpha-Release?

joshuamegnauth54:
Good news, everyone!

I may have fixed this issue. I am able to replicate the problem 100% of the time using the instructions @git-f0x posted above. Let me know in the PR if the patch worked for those who experienced the issue or else I can try further to fix it.

[...]

PR #249, opened 2024-04-29
fix: Properly cache non-custom wallpaper previews

joshuamegnauth54:
Closes: #211

The linked issue explains the problem and how to replicate it. Previews for wallpapers in the default folders were not cached which lead to blank previews.

I fixed it by retrieving the current wallpaper from `cosmic-bg` on load. This also fixes a similar issue with blank previews for colors.


mmstick:
Thanks!

PR #250, opened 2024-04-29
Themes! 🎉 

edfloreshz:
This pull request introduces a new Themes page within the desktop section, allowing users to effortlessly personalize their experience by switching between predefined themes.

I created a [repository](https://github.com/edfloreshz/cosmic-themes) with a pair of themes to start with, I'm wiling to transfer the repository to S76. This would allow theme submissions through pull requests, giving the team full control over what themes are included in the desktop environment.

screenshot-2024-04-29-08-30-27

screenshot-2024-04-29-08-43-24


mmstick:
Requires input from @pop-os/ux. I could see this being a sub-page of the appearance page.

TheESN:
I personally think that this page should be called colour schemes, since it doesn't actually change a COSMIC theme

edfloreshz:
@TheESN This actually makes sense, l'll make the required  changes today.

ryanabx:
Idea for future work: If themes have a standard configuration file (i.e. potentially a .ron file), support loading colors from a .ron file.

edfloreshz:
I was actually thinking of adding a button to save the current configuration as a new local theme.

ryanabx:
> > Idea for future work: If themes have a standard configuration file (i.e. potentially a .ron file), support loading colors from a .ron file.
> 
> There are buttons on the appearance page for importing and exporting themes

Oh! I didn't realize, neat!

> I was actually thinking of adding a button to save the current configuration as a new local theme.

Yes, that would make a lot of sense, assuming people want to share their themes as much as ~~hyprland~~ users share their dotfiles :sweat_smile:

edfloreshz:
Even if sharing is not a priority, it would be nice to let users save their current configuration without managing the filesystem directly.

ryanabx:
However, thinking about this more:

I don't think the theme.ron abstraction makes much sense, might make more sense to just keep them separate. That'd reduce the number of .ron files.

I also would like to reduce the number of git submodules used in projects, and 2 layers of recursion is a bit much for my liking (personal preference though)

Also, extending the configs-to-ron stuff, it's probably worth looking into importing/exporting for panel configurations, which is potentially the only other appearance/layout setting that'd be good to have import/export for. If no one got to it in like a month's time I'd probably look into it myself

edfloreshz:
The intention here is to provide a color scheme with both light and dark variants, otherwise both variants could not be grouped and would also require file nomenclature to detect variants.

Perhaps the sub-module approach could be improved.

ryanabx:
I think I'm concerned about introducing path dependencies with theme.ron depending on the light and dark files.

An alternative that I think would be better (if we want theme groupings) is just bundle all subvariants of a theme into the one theme.ron file.

edfloreshz:
Yeah it definitely makes sense and I agree with that point, I'm open to change but I have to say that I'm concerned about having huge files that are hard to read, if we find a way to remove path dependencies without having huge files then I'm all in!

maria-komarova:
We are ready to consider it for Cosmic v2, but most likely themes won't be in the Settings but on our website. But no definite plans until we get to v2.

edfloreshz:
I believe Settings would be the perfect place to have them, is this open to consideration?

A great workflow would be for users to install `.ron` files to their user directory from the website through the Settings app, then pick or delete their installed color schemes there.

TheESN:
Idea for colour schemes preview

изображение

1280px:
What if we move all the theme controls inside of submenu/subpage for each theme, and also separate them by dark/light mode instead of per-theme separation? 🤔

изображение

vtmx:
One question, from what I understand the system will do some calculations with the base colors to change parts of the interface elements. It will be possible to manually change part of these elements in some configuration file. For example, if I want to change the color of the buttons outside the context of the main colors, will it be possible?

wash2:
> One question, from what I understand the system will do some calculations with the base colors to change parts of the interface elements. It will be possible to manually change part of these elements in some configuration file. For example, if I want to change the color of the buttons outside the context of the main colors, will it be possible?

This is possible, but it's not customizable via the settings application.

TheESN:
> One question, from what I understand the system will do some calculations with the base colors to change parts of the interface elements. It will be possible to manually change part of these elements in some configuration file. For example, if I want to change the color of the buttons outside the context of the main colors, will it be possible?

I think it's possible via manually editing exported .ron file of theme

PR #257, opened 2024-05-01
fix(appearance): use same theme border radius for dock and panel.

edfloreshz:
This PR makes the Dock and Top Panel follow the corner radius specified by the theme.

@wash2

### DIFF
--- a/cosmic-settings/src/pages/desktop/appearance.rs
+++ b/cosmic-settings/src/pages/desktop/appearance.rs
@@ -73,10 +73,6 @@ pub struct Page {
     interface_text: ColorPickerModel,
     control_component: ColorPickerModel,
     roundness: Roundness,
-    dock_config_helper: Option<Config>,
-    dock_config: Option<CosmicPanelConfig>,
-    panel_config_helper: Option<Config>,
-    panel_config: Option<CosmicPanelConfig>,
 
     icon_theme_active: Option<usize>,
     icon_themes: IconThemes,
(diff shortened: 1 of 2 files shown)


edfloreshz:
I believe that padding will be necessary to make this look great, but other than that, this is pretty much done.

cosmic-settings/src/pages/desktop/appearance.rs
```diff
@@ -149,6 +154,17 @@ impl
                 && c != theme.palette.accent_yellow
         });
 
+        let panel_config_helper = CosmicPanelConfig::cosmic_config("Panel").ok();
```
wash2, line 157: I think it would be ok to get the panel and dock configs in the update_panel_radii method. Is there a that they are added to the Page?

cosmic-settings/src/pages/desktop/panel/inner.rs
```diff
@@ -502,11 +503,54 @@ impl PageInner {
         }
 
         if panel_config.anchor_gap || !panel_config.expand_to_edges {
-            panel_config.border_radius = 8;
+            let radii = Self::get_corner_radii().radius_xl[0] as u32;
```
wash2, line 506: `cosmic::theme::system_preference().cosmic().corner_radii` should work here. I don't think it requires a new method.

edfloreshz: Awesome! I'll make the change.

cosmic-settings/src/pages/desktop/appearance.rs
edfloreshz: No reason, fixed!

wash2 (approved):

PR #259, opened 2024-05-02
fix(panel): use radius_s for panel corner radii

ryanabx:


wash2 (changes requested):
Thanks, the appearance page also needs to be updated to set `radius_s`

ryanabx:
@wash2 Did I get all the spots now?

wash2:
Yup, thanks!

Issue #252, opened 2024-04-30
Desktop > Wallpaper: Missing button of added image on subsequent visits to the page


git-f0x:
When the Wallpaper page is first opened, the added image is present both on the list and in the large preview.
When going to another page and returning (without closing Settings), the added image is missing in the list ~~and the preview is black~~ (though both appear for a split second), until scrolling or something is hovered over.

https://github.com/pop-os/cosmic-settings/assets/150025636/3d6582c9-a5d9-43c3-9416-d587ababcd12

git-f0x:
The preview being blank is fixed, but the added wallpaper still doesn't show up on the list until something is interacted with on subsequent visits to the page.

joshuamegnauth54:
Sorry! I'm responding to all of this a bit late. :laughing: 

I attempted to fix this issue over the last few days but I haven't made much progress. The PR you commented on fixed a related issue (which was likely a regression I caused!) but doesn't do much for this issue.

I'll likely continue trying to fix this issue, but I'm definitely stumped for now. The custom wallpaper buttons are successfully created despite not appearing. It's weird.

git-f0x:
This seems to have been fixed (though there is sometimes a short flicker in the list when opening the page, probably because the custom wallpaper is loaded).

PR #264, opened 2024-05-03
fix: Regression leading to blank custom previews

joshuamegnauth54:
I've encountered intermittent blank previews for custom wallpapers while ~~working on~~ failing to fix a different bug, but after this patch I haven't seen any empty previews at all.

I originally intended to bundle this fix with the other fix, but I haven't made much progress with that bug.


git-f0x:
This might fix #252.

PR #269, opened 2024-05-04
chore: move add applets page to context drawer

edfloreshz:
This PR moves the add applets page to context drawer, removing the extra window.

@mmstick I couldn't find the multi-window feature flag anywhere.

### DIFF
--- a/cosmic-settings/src/app.rs
+++ b/cosmic-settings/src/app.rs
@@ -3,8 +3,7 @@
 
 use crate::config::Config;
 use crate::pages::desktop::{
-    self, appearance,
-    dock::{self, applets::ADD_DOCK_APPLET_DIALOGUE_ID},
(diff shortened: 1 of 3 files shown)


wash2:
iced-sctk is multi-window by default actually. The multi-window feature won't be needed when the settings app is switched to the winit backend though.

wash2 (approved):

ryanabx:
> iced-sctk is multi-window by default actually. The multi-window feature won't be needed when the settings app is switched to the winit backend though.

Does that mean we can now switch settings to winit?

That'd be exciting, because the scroll speed is currently so slow on the settings app only because of the backend being used

mmstick:
It would still be slow with winit until GPU rendering is enabled. There is a rendering glitch with the image button that prevents us from using wgpu.

Issue #299, opened 2024-05-09
[Displays] Don't allow disabling a display if it's the only active display


WatchMkr:
1. Setup two monitors
2. Disable one
3. It's possible to disable the second monitor. Instead, make "Enable display" insensitive it in this case.

PR #321, opened 2024-05-19
fix: Disallow disabling only active display

joshuamegnauth54:
Closes: #299

### DIFF
--- a/cosmic-settings/src/pages/display/mod.rs
+++ b/cosmic-settings/src/pages/display/mod.rs
@@ -469,12 +469,23 @@ impl Page {
                 .button_alignment(Alignment::Center)
                 .on_activate(Message::Display);
 
-            let display_enable = list_column().add(cosmic::widget::settings::item(
-                &*text::DISPLAY_ENABLE,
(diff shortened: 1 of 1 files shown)


mmstick (changes requested):
This check could be moved to the toggle display button. Buttons are insensitive when their `on_press` property is unset.

joshuamegnauth54:
The `Toggler` itself doesn't seem to have `on_press`. Should I wrap it in a `Button`?

mmstick:
No, by not assigning a message, it will use an insensitive style and be unclickable. There should be a `maybe` variant of the method which takes an `Option`.

joshuamegnauth54:
It looks like this now:

disabled_01
disabled_02
disabled_03

The first and third displays are disabled so the toggle still shows. The second display is the only enabled display so the toggle is hidden.

Alsooo I'm not sure why CI is failing. I rebased on master and force pushed too.

mmstick:
Seems we'll need to improve the toggler widget to support an insensitive variant without a message.

mmstick (approved):

PR #323, opened 2024-05-19
Match window corner radius switching to other components

### LINKED ISSUE #295: [Desktop > Appearance] Tweak app window corner radius switching
Low priority:
Currently, the app window corners use the same radius on Square and Slightly Round, and a different (larger) radius on Round.
The NavBar (and some other components), use the same radius on Round and Slightly Round, and a different radius on Square.
The app window corners should match the NavBar switching behavior, and use the same radius as Round on Slightly Round, while keeping the current radius on Square.
<sub>This was approved by Maria.</sub>

git-f0x:
Closes #295.
This makes the window corners change in sync with e.g. the NavBar.

### DIFF
--- a/cosmic-settings/src/pages/desktop/appearance.rs
+++ b/cosmic-settings/src/pages/desktop/appearance.rs
@@ -326,7 +326,7 @@ impl From<Roundness> for CornerRadii {
             },
             Roundness::SlightlyRound => CornerRadii {
                 radius_0: [0.0; 4],
-                radius_xs: [2.0; 4],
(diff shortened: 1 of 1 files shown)


mmstick:
Maybe the window corner radius should be changed in libcosmic

git-f0x:
That's what Maria suggested (setting window corners to @radius_s). Should I close this?
Or maybe if setting @radius_xs to 4 doesn't cause any issues, it can be added (since a radius of 2 is barely visible).

Issue #10, opened 2022-06-02
Settings > Power


maria-komarova:

maria-komarova:
The Power & Battery panel in COSMIC Settings is where users can adjust settings for the battery to either save power or increase the performance. 

![Power](https://user-images.githubusercontent.com/57502897/211644724-0ef192c4-7c83-46b1-a15f-803eaea6c6ce.png)

New addition to this panel will be cards with connected devices. This section provides users with an immediately recognizable overview of the connected devices and the state of their batteries. This section shouldn’t be present when no devices are connected. 

Another addition to the panel is a Battery Health section with the toggle to limit battery charging that mirrors the feature we included in the Power top panel applet.

Graphics mode settings have now been moved to the Displays panel based on the results of the usability testing we conducted. To make sure people still have a way to find the graphics mode from this panel and learn how it can help preserve power we included a link to the settings under the Related settings section at the bottom of the page.

The mockups and a few other details for the panel could be found in [this Figma file](https://www.figma.com/file/142SZP1yG2d34SUdDvkaYV/COSMIC-Settings---Developer-Handoff?node-id=693%3A55166).

maria-komarova:
Designs updated for the 1st release ([Figma file](https://www.figma.com/design/Z6Z9Ox6WNkiiM8gcAkcElS/COSMIC-Settings-First-Release---Developer-Handoff?node-id=9753-92877&m=dev)):
![Power   Battery panel](https://github.com/user-attachments/assets/be2a9e22-2861-408d-8ad4-f96e2c533e5f)

liljom:
Is it okay to comment or feature request here? (just delete it if not)
In other distros it's so nice to have setting for power mode when laptop charger is plugged/unplugged. Like: set Extended Battery Life mode when unplugged, set High Performance mode when plugged.

iamkartiknayak:
> Is it okay to comment or feature request here? (just delete it if not)

There's a related issue here #595

WatchMkr:
Additional features will be added in a later release.

chinoto:
I just started using this yesterday and there happened to be an update later the same day, which I installed, but I'm still not seeing an option for Power Button Behavior. I accidentally shutdown my laptop twice yesterday, but I managed to remember to do `Super, "suspend"` the other times.

If it helps, here are the versions of the packages I have. "obs_cosmic" is an alias for https://download.opensuse.org/repositories/X11:/COSMIC:/Next/openSUSE_Factory/
[...]

PR #326, opened 2024-05-21
Power page & power profiles

burakturkerdev:
I implemented power profiles to settings via power profiles daemon.

### DIFF
--- a/cosmic-settings/src/pages/power/mod.rs
+++ b/cosmic-settings/src/pages/power/mod.rs
@@ -1,0 +1,21 @@
+use cosmic_settings_page as page;
+
+pub mod profiles;
+
+#[derive(Default)]
+pub struct Page;
+
+// TO-DO implement localization after find out how it works 
+impl page::Page<crate::pages::Message> for Page {
+    fn info(&self) -> page::Info {
+        page::Info::new("power", "preferences-system-time-symbolic")
+            .title("Power & Sleep")
+            .description("Settings for power and sleep")
+    }
+}
+
+impl page::AutoBind<crate::pages::Message> for Page {
+    fn sub_pages(page: page::Insert<crate::pages::Message>) -> page::Insert<crate::pages::Message> {
+        page.sub_page::<profiles::Page>()
+    }
+}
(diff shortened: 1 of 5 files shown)


mmstick (changes requested):
Must use system76-power's dbus interface, rather than powerprofilesctl.

Drakulix:
> Must use system76-power's dbus interface, rather than powerprofilesctl.

Or implement what cosmic-applet-power is doing and pick either (preferring system76-power), depending on what is available. In general that code is a good reference for how this page could work.

ryanabx:
Like @mmstick said, these settings are controlled via a dbus interface. See: https://github.com/pop-os/cosmic-applets/blob/master/cosmic-applet-battery/src/backend/power_daemon.rs for the system76-power backend and https://github.com/pop-os/cosmic-applets/blob/master/cosmic-applet-battery/src/backend/power_profiles.rs for the power profiles daemon backend

ryanabx:
Thank you! In the future, this can be extended to use both backends, but for now I think this works fine

cosmic-settings/src/pages/power/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+    fn info(&self) -> page::Info {
+        page::Info::new("power", "battery-symbolic")
+            .title(fl!("power"))
+            .description(fl!("power", "desc"))
+    }
+}
+
+impl page::AutoBind<crate::pages::Message> for Page {
+    fn sub_pages(page: page::Insert<crate::pages::Message>) -> page::Insert<crate::pages::Message> {
+        page.sub_page::<profiles::Page>()
```
mmstick, line 20: This doesn't need to be a sub-page. It should be a section of this page. See the mockup here: https://github.com/pop-os/cosmic-settings/issues/10

mmstick (changes requested):
See

cosmic-settings/src/pages/power/mod.rs
burakturkerdev: I was planning implementing sleep/screen saver settings. And i thought structure should be like general power page and sub-pages.

Power
     -- Sleep
     -- Power Profiles
 
Won't be so complicated if we put all power related settings in same page? Because we can still add things to power profiles sub page like auto switch power saver mode below 10% battery etc.

burakturkerdev:
@mmstick Updated to section from page.

mmstick (approved):

mmstick:
Thanks!

PR #332, opened 2024-05-24
Add i18n(es): Add Spanish translation

Adrianowsky:
Add Spanish translation


Adrianowsky:
I translated the COSMIC settings into Spanish, my native language.

PR #342, opened 2024-05-27
Multiple backend support for power profiles

burakturkerdev:
In last commit, I implemented only the S76 backend for backend. Now, it searches for S76 backend and if doesn't find it, it automatically checks for the power profiles backend as well. If there is no backend, it displays a 'backend not found' message instead of showing the power profiles section.


i18n/en/cosmic_settings.ftl
```diff
@@ -452,3 +452,4 @@ power-profiles = Power Profiles
   .performance-desc = Maximum performance but high power consumption.
   .balanced-desc = Balanced performance and power consumption.
   .battery-desc = Low performance but low power consumption.
+  .nobackend = Backend not found. Install S76Power or PowerProfiles.
```
mmstick, line 455: It should say system76-power instead of S76Power.

cosmic-settings/src/pages/power/backend/mod.rs
```diff
@@ -53,56 +85,155 @@ impl PowerProfile {
...
+            Err(e) => {
+                println!("[Cosmic Settings] Problem while setting power profile.");
+                return;
+            }
+        };
+
+        match profile {
+            PowerProfile::Performance => match daemon.performance().await {
+                Ok(x) => println!("[Cosmic Settings] Performance mode activated."),
+                Err(e) => println!("{e}"),
```
mmstick, line 113: Use `tracing::error` or `tracing::warn` instead of println

cosmic-settings/src/pages/power/backend/mod.rs
```diff
@@ -102,23 +102,23 @@ impl SetPowerProfile for S76Backend {
         let daemon = match get_s76power_daemon_proxy().await {
             Ok(c) => c,
             Err(e) => {
-                println!("[Cosmic Settings] Problem while setting power profile.");
+                tracing::error!("[Cosmic Settings] Problem while setting power profile.");
```
mmstick, line 105: It's not necessary to add `[Cosmic Settings]`.

mmstick (approved):

PR #351, opened 2024-05-31
Fix centering displays in arrangement

jasonrhansen:
I don't know what the reason was to multiply the max dimensions by 1.25, but it causes the displays to not be centered properly in the arrangement section for me.

This is what it looks like for me when multiplying by 1.25
screenshot-2024-05-31-22-21-19

This is what it looks like when I remove it
screenshot-2024-05-31-22-25-19

### DIFF
--- a/cosmic-settings/src/pages/display/arrangement.rs
+++ b/cosmic-settings/src/pages/display/arrangement.rs
@@ -124,8 +124,8 @@ impl<'a, Message: Clone> Widget<Message, cosmic::Theme, Renderer> for Arrangemen
 
         let state = tree.state.downcast_mut::<State>();
         state.max_dimensions = (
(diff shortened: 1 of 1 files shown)


mmstick:
Are you able to comfortably drag display 1 to the left of display 0?

jasonrhansen:
With or without my change dragging a display from one side to the other of another display sometimes requires me to scroll.

jasonrhansen:
Before my change, there was more empty space on the left than the right to drag to. This meant there wasn't enough space to drag the large display 0 to the right of display 1 without multiple steps. After my changes there is equal room on each side and I can drag either monitor to either side with enough space to fit.

jasonrhansen:
But either way it can be a little awkward dragging displays because of the fact that it doesn't receive mouse events when dragging outside of the bounds of the scroll container.

mmstick (approved):

PR #352, opened 2024-06-01
Match text on Power page to design

git-f0x:
Also includes a minor grammar fix for an error message.


cosmic-settings/src/pages/power/mod.rs
```diff
@@ -85,7 +85,7 @@ fn profiles() -> Section<crate::pages::Message> {
                     section = section.add(item);
                 }
             } else {
-                let item = widget::Text::new(fl!("power-profiles", "nobackend"));
+                let item = widget::Text::new(fl!("power-mode", "nobackend"));
```
mmstick, line 88: This should be `widget::text::body`

git-f0x: I'm not very familiar with this, but when compiling, it says that the `body` function isn't found in module `widget::text`. Any tips as to what should be imported instead of `cosmic::widget::text`?

mmstick: You can see available text variants here: https://pop-os.github.io/libcosmic/cosmic/widget/text/index.html
The body function does exist here.

mmstick: This module is importing `use cosmic::iced::widget;` when it should be importing `cosmic::widget`

git-f0x: It seems to compile if a change that line to `cosmic::widget::text::body(...)`, but I'm likely not importing something correctly.

mmstick: Another issue is that the update method is blocking the runtime, when it should be spawning commands.

git-f0x: I'll push the variant that compiles so that the rest of the changes can be reviewed.

cosmic-settings/src/pages/power/mod.rs
```diff
@@ -85,7 +85,7 @@ fn profiles() -> Section<crate::pages::Message> {
                     section = section.add(item);
                 }
             } else {
-                let item = widget::Text::new(fl!("power-mode", "nobackend"));
+                let item = cosmic::widget::text::body(fl!("power-mode", "nobackend"));
```
mmstick, line 88: Need to import `cosmic::widget` instead of `cosmic::iced::widget`

cosmic-settings/src/pages/power/mod.rs
git-f0x: Oh wow, your responses didn't show up at all until the site was refreshed. Sorry for seemingly ignoring your answers.

git-f0x: >Another issue is that the update method is blocking the runtime, when it should be spawning commands.

What should I do about this?

mmstick: The update method needs to be restructured to return `cosmic::command`s instead of blocking.

git-f0x: I'm not sure that I'm capable of doing that. :(

mmstick: Then I'll do it once this is merged

mmstick (approved):

mmstick:
I'm probably going to remove the backend code anyway so that it uses the same code used by the battery applet.

PR #377, opened 2024-06-12
Adding dialog to confirm display changes.

### LINKED ISSUE #365: [Feature Request] Add a "Confirm" prompt to Display changes
Switching the resolution or refresh rate can render (ha) them unusable.  There should be a confirmation prompt with a countdown that reverts the display to the previous setting.

pknepps:
Resolves #365 
This feature adds a confirmation prompt with a countdown once a setting is changed. The settings will automatically revert if the timer runs to 0.

The dialog will appear for orientation or scale changes, and the enabling/disabling of a display. It is currently not implemented for display position, resolution, nor refresh rate changes. For any setting change in which someone wishes to add the dialog, I have added a macro to make implementing the dialog simple.

I am not sure if the design choice I made for closing the dialog after timeout was a good one. I made a Widget which just sends a continuous stream of a single message. If that was a good choice, then I am not sure if it is okay to be a private struct at the bottom of the file, or if it should be public or have its own file somewhere else within the repository.

### DIFF
--- a/cosmic-settings/src/pages/display/mod.rs
+++ b/cosmic-settings/src/pages/display/mod.rs
@@ -6,19 +6,18 @@ pub mod arrangement;
 
 use crate::{app, pages};
 use arrangement::Arrangement;
-use cosmic::iced::{Alignment, Length};
+use cosmic::iced::{time, Alignment, Length};
 use cosmic::iced_widget::scrollable::{Direction, Properties, RelativeOffset};
 use cosmic::prelude::CollectionWidget;
 use cosmic::widget::{
-    column, container, dropdown, list_column, segmented_button, tab_bar, toggler,
+    self, column, container, dropdown, list_column, segmented_button, tab_bar, toggler,
 };
 use cosmic::{command, Apply, Command, Element};
 use cosmic_randr_shell::{List, Output, OutputKey, Transform};
 use cosmic_settings_page::{self as page, section, Section};
 use slab::Slab;
 use slotmap::{Key, SlotMap};
-use std::collections::BTreeMap;
-use std::{process::ExitStatus, sync::Arc};
+use std::{collections::BTreeMap, process::ExitStatus, sync::Arc};
 
 /// Display color depth options
 #[derive(Clone, Copy, Debug)]
@@ -61,6 +60,12 @@ pub enum Message {
     ColorDepth(ColorDepth),
     /// Set the color profile of a display.
     ColorProfile(usize),
+    /// The dialog was cancelled, and will revert settings.
+    DialogCancel,
+    /// The dialog was completed.
+    DialogComplete,
+    /// How long until the dialog automatically cancelles, in seconds.
+    DialogCountdown,
     /// Toggles display on or off.
     DisplayToggle(bool),
     /// Configures mirroring status of a display.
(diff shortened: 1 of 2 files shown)


mmstick (changes requested):
I'm not sure why, but I'm not seeing a dialog appear from this branch.

Either way, it would be better to use Commands for handling the countdown, instead of the ContinuousMessageStream. When setting the dialog, add this command to be returned from the update method

```rs
command::future(async {
    tokio::time::sleep(Duration::from_secs(1)).await
    Message::DialogCountdown(10)
})
```

You can decrement the value by 1 until it reaches 0, and use the value as the countdown in the dialog.

```rs
Message::DialogCountdown(countdown) => {
    if countdown == 0 {
        self.dialog = None;
    } else {
        self.dialog_countdown(countdown);
        return command::future(async move {
            tokio::time::sleep(Duration::from_secs(1)).await
            Message::DialogCountdown(countdown - 1)
        });
    }
}
```

cosmic-settings/src/pages/display/mod.rs
```diff
@@ -115,6 +129,30 @@ pub struct Page {
...
+///     let revert_action = Randr::Scale(self.config.scale);
+///     set_dialog!(self, revert_action);
+///
+///     self.config.scale = scale;
+///     self.exec_randr(Randr::Scale(scale))
+/// }
+/// ```
+macro_rules! set_dialog {
+    ($self: ident, $revert_action: expr) => {
+        $self.dialog = Some(($revert_action, time::Instant::now()))
```
mmstick, line 154: I think it'd be better to make this a method of our Page

cosmic-settings/src/pages/display/mod.rs
```diff
@@ -6,19 +6,30 @@ pub mod arrangement;
...
 };
 use cosmic::{command, Apply, Command, Element};
 use cosmic_randr_shell::{List, Output, OutputKey, Transform};
 use cosmic_settings_page::{self as page, section, Section};
 use slab::Slab;
 use slotmap::{Key, SlotMap};
-use std::collections::BTreeMap;
-use std::{process::ExitStatus, sync::Arc};
+use std::thread::current;
+use std::{collections::BTreeMap, process::ExitStatus, sync::Arc};
```
mmstick, line 32: Make sure to remove unused imports

cosmic-settings/src/pages/display/mod.rs
pknepps: I changed it to a method with the signature:
```rust
fn set_dialog(&mut self, revert_request: Randr, current_request: Randr) => Command<app::Message> {
```

cosmic-settings/src/pages/display/mod.rs
pknepps: I have done so.

pknepps:
I have completed the changes to the best of my ability. 

As to the dialog not appearing, I attempted to try to recreate the issue on a VM, but had trouble getting COSMIC to work at all. Let me know if you still have don't see the dialog and I'll spend more time trying to figure it out.

mmstick:
Squashed all the commits and fixed the remaining warnings. It looks good enough to merge now. Thanks!

PR #389, opened 2024-06-30
Display input speed (mouse + touchpad) range as [1,100]

### LINKED ISSUE #285: [Input Device > Mouse/Touchpad] Normalize the speed range
Mouse and touchpad speed were correctly modified to use 10 as the lowest value and 80 as the highest value for the libinput speed range. Let's normalize the UI to 1-100 so it makes more sense to users and matches other 1-100 values.

mttmartin:
Display input speed (mouse + touchpad) range as [1,100]. The minimum and maximum values of the selectable range should result in the same value being passed to libinput as before.

Fixes #285


mmstick (approved):

mmstick:
Thanks!

PR #409, opened 2024-07-18
feat(displays): pan left and right when dragging displays to edge

mmstick:
Improves the experience when dragging displays around the display arrangement widget


Drakulix (approved):
Works nicely and good looks good, though I wonder if we should also support panning up or down for vertical configurations?

In any case this is much better than the current status quo.

wash2 (approved):

Issue #439, opened 2024-08-01
Updated Desktop page with sub-categories


maria-komarova:
Desktop page has been updated because the new Window management page now includes Super key action and window control options. So Panel and Dock pages move up in the information architecture of the application. Less steps for people to get to those. 

![Desktop](https://github.com/user-attachments/assets/08dc9e12-4287-4aa6-aa24-fe954aa3b35d)

[Figma file](https://www.figma.com/design/Z6Z9Ox6WNkiiM8gcAkcElS/COSMIC-Settings-First-Release---Developer-Handoff?node-id=13834-11011&t=cagN4rG2n71Xfuj0-1).

PR #448, opened 2024-08-02
[Desktop] Rearrange pages to match design

git-f0x:
This matches the design in #439. Draft because desc for window-management is reported as not existing for some reason, so help on that would be welcome.
Screenshot without Window Management description:
screenshot-2024-08-02-19-23-07

Edit: Unrelated, but the non-scrolling page title disappears when the window is shrunk vertically, instead of staying and a scrollbar appearing, which results in the Desktop page with the extra entries opening without a title at the default window size.


i18n/en/cosmic_settings.ftl
```diff
@@ -233,6 +224,20 @@ x-hours = { $number ->
     *[other] { $number } hours
 }
 
+## Desktop: Window management
+
+window-management = Window management
```
wash2, line 229: There is already a key for window-management on line 69, so you can add the new description there.

git-f0x: Yeah I just noticed. I added -appearance to the one in Appearance. Let me know if it should be done differently.

git-f0x: Also should hot-corner be removed, or is it planned to be added at some point?

wash2: I think it is ok to leave it in for now

git-f0x:
This might need some tiny code styling changes, since things aren't alphabetically arranged in some places.

wash2:
> This might need some tiny code styling changes, since things aren't alphabetically arranged in some places.

Ya that would be good

git-f0x:
Since the Desktop and Panel page was removed, this might create issues with Launcher searching "Panel" on existing installs, since the non-functional desktop entry for Desktop and Panel remains.

wash2:
https://github.com/pop-os/cosmic-settings/blob/master/debian/install also needs to be updated to remove the old file and add the new ones.

git-f0x:
It seems the icons for the subpages are larger in the designs, but not sure how to change that, since [this function](https://github.com/pop-os/cosmic-settings/blob/34d2b82376594e689ced2bcf8a23e7be0167bf06/page/src/lib.rs#L110) doesn't seem to take any size.

PR #458, opened 2024-08-04
Display refresh rates as integers

tobyrg:
Display the refresh rates as integers to match design.

Before:
image

After:
image

### DIFF
--- a/cosmic-settings/src/pages/display/mod.rs
+++ b/cosmic-settings/src/pages/display/mod.rs
@@ -1109,7 +1109,7 @@ pub fn display_configuration() -> Section<crate::pages::Message> {
 fn cache_rates(cached_rates: &mut Vec<String>, rates: &[u32]) {
     *cached_rates = rates
         .iter()
-        .map(|&rate| format!("{:>3}.{:02} Hz", rate / 1000, rate % 1000))
+        .map(|&rate| format!("{} Hz", rate / 1000))
         .collect();
 }
 
--- a/cosmic-settings/src/pages/display/mod.rs
+++ b/cosmic-settings/src/pages/display/mod.rs
@@ -1109,7 +1109,7 @@ pub fn display_configuration() -> Section<crate::pages::Message> {
 fn cache_rates(cached_rates: &mut Vec<String>, rates: &[u32]) {
     *cached_rates = rates
         .iter()
-        .map(|&rate| format!("{} Hz", rate / 1000))
+        .map(|&rate| format!("{} Hz", (rate as f32 / 1000.0).round() as u32))
         .collect();
 }
 


tobyrg:
Only issue I have noticed is that there are duplicates on some monitors which I think an issue in cosmic-randr.

Screenshot:
image

cosmic-randr output:
```
DP-2 (enabled)
  Model: 24G2W1G4
  Physical Size: 530 x 300 mm
  Position: 0,0
  Scale: 100%
  Transform: normal
  Adaptive Sync: false

  Modes:
    1920x1080 @ 144.000 Hz (current)
    1920x1080 @ 120.000 Hz
    1920x1080 @ 119.982 Hz
    1920x1080 @ 119.880 Hz
    1920x1080 @  99.930 Hz
    1920x1080 @  74.973 Hz
    1920x1080 @  60.000 Hz (preferred)
    1920x1080 @  59.940 Hz
    1920x1080 @  50.000 Hz
    1680x1050 @  60.000 Hz
     1440x900 @  60.000 Hz
    1280x1024 @  75.025 Hz
    1280x1024 @  60.020 Hz
     1280x800 @  60.000 Hz
     1280x720 @  60.000 Hz
     1280x720 @  59.940 Hz
     1280x720 @  50.000 Hz
     1024x768 @ 119.926 Hz
     1024x768 @  99.991 Hz
     1024x768 @  75.029 Hz
     1024x768 @  70.069 Hz
     1024x768 @  60.004 Hz
      832x624 @  74.551 Hz
      800x600 @ 119.933 Hz
      800x600 @  99.862 Hz
      800x600 @  75.000 Hz
      800x600 @  72.188 Hz
      800x600 @  60.317 Hz
      800x600 @  56.250 Hz
      720x576 @  50.000 Hz
      720x480 @  60.000 Hz
      720x480 @  59.940 Hz
      720x400 @  70.082 Hz
      640x480 @ 119.803 Hz
      640x480 @  99.825 Hz
      640x480 @  75.000 Hz
      640x480 @  72.809 Hz
      640x480 @  66.667 Hz
      640x480 @  60.000 Hz
      640x480 @  59.940 Hz
      ```

**Comment by Drakulix** [comment]
While it would be nice to clean up the output a bit, these are actually two different modes and not duplicates:
```
1920x1080 @ 119.982 Hz
1920x1080 @ 119.880 Hz
```

For this reason I don't think we should truncate viable information.

tobyrg:
I noticed that when comparing to KDE Plasma, the settings only display a single 120 Hz option in the dropdown. This looks cleaner to me, and I was wondering how similar refresh rates should be handled but if it's preferred to show all possible refresh rates in the settings should I close this PR then?

iotku:
I think from a UX point of view chances are if I'm looking to change my refresh rate to something higher than default for gaming I just want the biggest number that my monitor supports so rounding up in that case and ignoring duplicates that are fractionally less is mostly reasonable behavior. (At a glance this appears to be how Plasma operates.)

On the other hand, for some specific applications (or perhaps finicky displays) a more precise refresh rate might be required. Theoretically, when dealing with NTSC video for instance someone might actually want their display closer to 59.94hz (Although the NTSC standard is *basically* 60Hz, historical reasons a lot of content is effectively 59.94hz and some displays explicitly support that refresh rate as well as 60Hz).

If there was a checkbox toggle or setting for "Advanced Display Settings" or something around those lines which showed all of the available modes that could be a decent middle ground.

Past that if some display drivers allow completely custom modes you wouldn't want those to be removed from the list because a user could explicitly have setup a custom mode for their particular use case, though I imagine setting a custom display mode from within cosmic-settings might be out of scope.

tobyrg:
I'm quite new to Rust, so I'm not sure what's the best way to deal with duplicate and make sure that the highest refresh rate out of the similar ones is in the refresh rate list.

PR #510, opened 2024-08-20
Update cosmic_settings.ftl

lucamosca1:
Attempt to merge, from english lang file all the missing string as placeholders. Try to keep existing italian ones (manually or via home-made script is a bit hard, see https://github.com/pop-os/cosmic-epoch/issues/756 )

### DIFF
--- a/i18n/it/cosmic_settings.ftl
+++ b/i18n/it/cosmic_settings.ftl
@@ -4,9 +4,47 @@ unknown = Sconosciuto
 
 number = { $number }
 
+## Networking: Wired
+
+wired = Via cavo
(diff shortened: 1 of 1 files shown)


mmstick (approved):

mmstick (changes requested):
This seems to be replacing a lot of existing translations with English, and adding English translations where they already exist.

lucamosca1:
Ok, just a question. Italian localizations are a lot older and not updated with english ones (588 lines against 411) and also in a different order. Is there a smarter way to handle the differences and merge the files even with a different order? Otherwise I have to rewrite the file manually. Thanks

PR #524, opened 2024-08-27
displays: Add X11 scaling settings

Drakulix:
~~Depends on https://github.com/pop-os/cosmic-comp/pull/779 being merged first.~~ cosmic-comp changes merged.

### DIFF
--- a/cosmic-settings/src/pages/display/mod.rs
+++ b/cosmic-settings/src/pages/display/mod.rs
@@ -10,15 +10,17 @@ use cosmic::iced::{time, Alignment, Length};
 use cosmic::iced_widget::scrollable::{Direction, Properties, RelativeOffset};
 use cosmic::prelude::CollectionWidget;
 use cosmic::widget::{
-    self, column, container, dropdown, list_column, segmented_button, tab_bar, toggler,
+    self, column, container, dropdown, list_column, segmented_button, tab_bar, text, toggler,
 };
 use cosmic::{command, Apply, Command, Element};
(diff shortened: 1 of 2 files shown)


cosmic-settings/src/pages/display/mod.rs
```diff
@@ -1098,6 +1127,43 @@ pub fn display_configuration() -> Section<crate::pages::Message> {
...
+    Section::default()
+        .title(fl!("legacy-applications"))
+        .descriptions(descriptions)
+        .view::<Page>(move |_binder, page, section| {
+            let descriptions = &section.descriptions;
+            widget::settings::view_section(&section.title)
+                .add(widget::settings::item_row(vec![widget::radio(
+                    widget::column()
+                        .push(text::heading(&descriptions[system]))
+                        .push(text(&descriptions[system_desc])),
```
git-f0x, line 1147: Looking at the designs, this should probably be `text::body` and `text::caption` for system and system_desc (and native/native_desc), respectively.

Drakulix: Totally right, thanks for pointing that out.

git-f0x:
I'm on my phone currently, so I can't do a proper comment, but line 1088 should use `space_xs`, and the `.push(...)` at line 1123 should be moved to an `if page.list.outputs.len() == 1` just above, in order to close #528.

Drakulix:
https://github.com/pop-os/cosmic-settings/issues/528 should be addressed by the latest changes.

mmstick (approved):

PR #534, opened 2024-08-30
Update dependencies

git-f0x:
Update to use the custom radio widget, as well as rearrange some things to be alphabetical.

### DIFF
--- a/cosmic-settings/src/pages/desktop/workspaces.rs
+++ b/cosmic-settings/src/pages/desktop/workspaces.rs
@@ -3,10 +3,9 @@
 
 // TODO make settings work
 
-use cosmic::widget::settings;
 use cosmic::{
     cosmic_config::{self, ConfigGet, ConfigSet},
-    widget::radio,
+    widget::{radio, settings, text},
     Apply, Element,
 };
 use cosmic_comp_config::workspace::{WorkspaceConfig, WorkspaceLayout, WorkspaceMode};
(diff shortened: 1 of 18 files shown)


cosmic-settings/src/pages/desktop/workspaces.rs
```diff
@@ -136,14 +135,14 @@ fn multi_behavior() -> Section<crate::pages::Message> {
             let descriptions = &section.descriptions;
             settings::view_section(&section.title)
                 .add(settings::item_row(vec![radio(
-                    &descriptions[span],
+                    text(&descriptions[span]),
```
mmstick, line 138: Make sure to use `text::body` instead

git-f0x: Oh, I've just noticed that workspace orientation didn't need changes. Should I maybe put just &*descriptions instead of using text?

mmstick: Sure

git-f0x: Sorry for the many questions, but a few places in touchpad.rs have `text(&description[])`: should that be changed to `&*description[]`, or would it be better for everything to use text::body?

mmstick: It's generally best to use `text::body` to ensure that the correct text style is set.

git-f0x: Should I also do this for other pages (e.g. Mouse), or just these two (and maybe import `text::body` instead of `text` where possible)?

mmstick: If you see any area that omitted it, it would be good to fix it.

git-f0x: I think I've caught most things, and nothing seems to be changed visually.

PR #540, opened 2024-08-30
feat(display): display as integer

wiiznokes:
same as #458, but will keep the decimal if there are two rounded values that are equals


mmstick (approved):

mmstick:
Thanks!

PR #542, opened 2024-08-31
Update cosmic_settings.ftl

lucamosca1:
added missing week days

### DIFF
--- a/i18n/en/cosmic_settings.ftl
+++ b/i18n/en/cosmic_settings.ftl
@@ -541,10 +541,13 @@ time-format = Date & Time Format
     .show-seconds = Show seconds
     .first = First day of week
     .show-date = Show Date on Top Panel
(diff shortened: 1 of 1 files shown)


mmstick (approved):

i18n/en/cosmic_settings.ftl
```diff
@@ -541,10 +541,13 @@ time-format = Date & Time Format
...
     .show-date = Show Date on Top Panel
-    .friday = Friday
-    .saturday = Saturday
     .sunday = Sunday
     .monday = Monday
+    .tuesday = Tuesday
+    .wednesday = Wednesday
+    .thursday = Thursday
+    .friday = Friday
+    .saturday = Saturday
```
mmstick, line 550: None of these are used right now, so it's not needed

PR #553, opened 2024-09-03
Feat/power connected devices

acolombier:
Add the "Connected device" section, as per the [Figma design](https://www.figma.com/design/142SZP1yG2d34SUdDvkaYV/COSMIC-Settings---Developer-Handoff?node-id=693-55166&t=69OybxpixJYCuIqi-1)

## Example

Screenshot from 2024-09-03 11-04-43
Screenshot from 2024-09-03 11-03-23
image

~Depends of https://github.com/pop-os/dbus-settings-bindings/pull/17~
Partly addresses #10

### DIFF
--- a/cosmic-settings/src/pages/power/backend/mod.rs
+++ b/cosmic-settings/src/pages/power/backend/mod.rs
@@ -1,5 +1,7 @@
 use chrono::Duration;
 use futures::FutureExt;
+use futures::future::join_all;
+use upower_dbus::{BatteryType, DeviceProxy};
 use zbus::Connection;
 
 mod ppdaemon;
@@ -239,6 +241,13 @@ pub struct Battery {
     pub remaining_time: String,
 }
 
+#[derive(Default, Debug, Clone)]
+pub struct ConnectedDevice {
+    pub model: String,
+    pub device_icon: &'static str,
+    pub battery: Battery,
+}
+
 async fn get_device_proxy<'a>() -> Result<upower_dbus::DeviceProxy<'a>, zbus::Error> {
     let connection = match Connection::system().await {
         Ok(c) => c,
@@ -252,6 +261,42 @@ async fn get_device_proxy<'a>() -> Result<upower_dbus::DeviceProxy<'a>, zbus::Er
         Ok(p) => p.get_display_device().await,
         Err(e) => Err(e),
     }
+}
+
+async fn enumerate_devices<'a>() -> Result<Vec<upower_dbus::DeviceProxy<'a>>, zbus::Error> {
+    let connection = match Connection::system().await {
+        Ok(c) => c,
+        Err(e) => {
+            tracing::error!("zbus connection failed. {e}");
+            return Err(e);
+        }
+    };
+
+    let devices = upower_dbus::UPowerProxy::new(&connection)
+        .await?
+        .enumerate_devices()
+        .await?;
+
+    let devices = futures::future::join_all(
+        devices
+            .into_iter()
+            .map(|path| DeviceProxy::new(&connection, path)),
+    )
+    .await;
+
+    let errors = devices.iter().filter(|device|device.is_err());
+    if errors.count() > 0 {
+        let mut errors: Vec<zbus::Error> = devices.into_iter().filter_map(std::result::Result::err).collect();
+        if errors.len() > 1 {
+            eprintln!("Multiple errors occurs when fetching connected device: {errors:?}. Only the last one will be returned.");
+            
+        }
+        return Err(errors.pop().unwrap());
+    }
+    Ok(devices
+        .into_iter()
+        .filter_map(std::result::Result::ok)
+        .collect())
 }
 
 async fn get_on_battery_status() -> Result<bool, zbus::Error> {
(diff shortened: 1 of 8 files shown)


cosmic-settings/src/pages/power/mod.rs
```diff
@@ -38,18 +44,43 @@ impl page::Page<crate::pages::Message> for Page {
...
+                //         model: "Test device 2".to_owned(),
+                //         device_icon: "laptop-symbolic",
+                //         battery: Battery::default()
+                //     },
+                //     ConnectedDevice {
+                //         model: "Test device 3".to_owned(),
+                //         device_icon: "network-wired-symbolic",
+                //         battery: Battery::default()
+                //     },
+                // ];
```
acolombier, line 70: I will remove this before merging, but thought I will keep it in during review to help with testing!

cosmic-settings/src/pages/power/mod.rs
```diff
@@ -86,14 +120,86 @@ fn battery_info() -> Section<crate::pages::Message> {
...
+                    let battery_percent =
+                        widget::text(format!("{}%", connected_device.battery.percent));
+
+                    // FIXME is that missing from the design?
+                    // let battery_time =
+                    //     widget::text(if connected_device.battery.remaining_duration > TimeDelta::zero() {
+                    //         &connected_device.battery.remaining_time
+                    //     } else {
+                    //         ""
+                    //     });
```
acolombier, line 154: @maria-komarova I believe you looked at the original design. Is this on purpose or an oversight?

cosmic-settings/src/pages/power/backend/mod.rs
```diff
@@ -270,81 +315,149 @@ async fn get_on_battery_status() -> Result<bool, zbus::Error> {
...
+            BatteryType::Network => "network-wired-symbolic",
+            BatteryType::Headset => "audio-headset-symbolic",
+            BatteryType::Speakers => "speaker-symbolic",
+            BatteryType::Headphones => "audio-headphones-symbolic",
+            BatteryType::Video => "video-display-symbolic",
+            BatteryType::OtherAudio => "audio-speakers-symbolic",
+            BatteryType::Printer => "printer-network-symbolic",
+            BatteryType::Scanner => "scanner-symbolic",
+            BatteryType::Camera => "camera-photo-symbolic",
+            _ => "bluetooth-symbolic",
```
acolombier, line 436: A few devices don't have an [icon](https://www.figma.com/design/yFxlLCtVYLHMuHBNSXLf7O/Icon-theme?node-id=514-3593&node-type=SYMBOL&t=5S42FVRN9kxzClNA-0). Should I create an issue?

cosmic-settings/src/pages/power/backend/mod.rs
```diff
@@ -270,81 +315,149 @@ async fn get_on_battery_status() -> Result<bool, zbus::Error> {
 }
 
 impl Battery {
-    pub async fn update_battery() -> Self {
-        let proxy = get_device_proxy().await;
+    pub async fn from_device(proxy: DeviceProxy<'_>) -> Self {
```
acolombier, line 318: Ideally, I would like to implement the From trait, but I have no idea how to do so with an async function. Any suggestion? Should I keep it like this?

cosmic-settings/src/pages/power/backend/mod.rs
```diff
@@ -270,81 +315,149 @@ async fn get_on_battery_status() -> Result<bool, zbus::Error> {
...
+        if let Ok(proxy) = proxy {
+            return Self::from_device(proxy).await;
         }
 
         Battery::default()
     }
 }
+
+impl ConnectedDevice {
+    async fn from_device_maybe(proxy: DeviceProxy<'_>) -> Option<Self> {
```
acolombier, line 402: Ditto From trait

cosmic-settings/src/pages/power/backend/mod.rs
mmstick: You cannot do that since it's not an async trait method.

cosmic-settings/src/pages/power/mod.rs
git-f0x: Also, the order should possibly be **battery icon - percentage**, instead of **percentage - battery icon**, to match the system battery info above, and possibly prevent the battery icon moving when the percentage changes.

acolombier: Yeah, I thought so too. Currently I have followed the design 1:1, but happy to change if there is an issue with the design

git-f0x:
The designs don't have the remaining battery time for connected devices (e.g. 2 hours left).

maria-komarova:
If we can have an accurate estimate, then we can add it.

git-f0x:
Would it possibly be better to flip the order of the battery icon and percentage for connected devices (icon left - text right), to match the system battery info above, and avoid the icon moving when the text changes?

acolombier:
Thanks for the mock-up @maria-komarova 

After investigation, it looks like UPower is always returning a null time estimation with my devices so I have changed the code to follow the new mock-up, in case UPower was to return a valid time.
(Note that currently the estimation string follow the template `hh:mm:ss left` instead of the more friendly form suggested, but this issue is also impacting the `Battery` section since this is reusing the same code. Happy to look at this in subsequent PR)

Here is the new look
image
@git-f0x are you happy with that new layout?

acolombier:
> whether this is merged doesn't depend on me

Oh I know, I was just asking so I could resolve the thread above :) 

> the system battery percentage doesn't seem to aligned properly

Good eyes - I'll make another PR, so I can also fix the time remaining label

mmstick (approved):

acolombier:
I added #555, would be great to merge it first, so I can update this one to use the same helper for remaining text generation.

acolombier:
I have updated the PR with the #555 changes - if CI passes, this is ready to go @mmstick

mmstick (approved):

mmstick:
Thanks!

PR #555, opened 2024-09-03
Fix battery remaining time alignment and enhance copy

acolombier:
# Actual

image

# Design mock

image

### DIFF
--- a/cosmic-settings/src/pages/power/mod.rs
+++ b/cosmic-settings/src/pages/power/mod.rs
@@ -3,8 +3,7 @@ mod backend;
 use self::backend::{GetCurrentPowerProfile, SetPowerProfile};
 use backend::{Battery, PowerProfile};
 
-use chrono::TimeDelta;
-use cosmic::iced::Length;
+use cosmic::iced::{Length, Alignment};
 use cosmic::iced_widget::row;
 use cosmic::widget::{self, column, radio, settings, text};
 use cosmic::Apply;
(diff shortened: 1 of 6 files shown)


cosmic-settings/src/pages/power/mod.rs
```diff
@@ -81,16 +80,16 @@ fn battery_info() -> Section<crate::pages::Message> {
...
+            let battery_time = text::body(page.battery.remaining_time());
 
             column::with_capacity(2)
                 .spacing(8)
                 .push(text::heading(&section.title))
-                .push(row!(battery_icon, battery_percent, battery_time).spacing(8))
+                .push(
+                    row!(battery_icon, battery_percent, battery_time)
+                        .align_items(Alignment::Center)
+                        .spacing(8),
```
git-f0x, line 91: Looking at the designs, the spacing here should possibly be 4. The column spacing also seems smaller in the designs, but not sure exactly what value.

maria-komarova:
Spacing between icon and copy should be @space_xxxs (4) and spacing between header (Battery) and the copy @space_xxs (8), same as between the headers and ListViews in other sections.

acolombier:
Apologies, I didn't know there was spacing value already available in the theme!

> Spacing between icon and copy should be @space_xxxs (4) and spacing between header (Battery) and the copy @space_xxs (8), same as between the headers and ListViews in other sections.

After double-checking the design, it looks like there is no spacing between the section header (Battery), and the icon+copy, unlike the other section, which have a 8px space between header and content.

image

I have updated to match the design, but let me know if it should have a @space_xxs (8) spacing as well.

image

mmstick (approved):

mmstick:
Thanks!

PR #557, opened 2024-09-04
feat(Workspace): add missing settings and update design

acolombier:
This work is mostly focused on the look-n-feel.

Note there is currently [a limitation impacting SVG rendering](https://github.com/iced-rs/iced/issues/2566) in iced 0.12.x, but this appear to be fixed in 0.13

# Design

image

# Current look

https://github.com/user-attachments/assets/880ccea8-b2d4-40af-9aea-752be378b57d


maria-komarova:
On the video it looks like the illustrations for workspaces spanning displays and being separate is very dark. Do you know why that it? Are those just SVGs?

I also can't tell from the video is workspace thumbnails change position on the illustration depending on their placement (left, right etc.)

acolombier:
> On the video it looks like the illustrations for workspaces spanning displays and being separate is very dark. Do you know why that it? Are those just SVGs?

Yes, this is due to the bug in iced I attached above. 

> I also can't tell from the video is workspace thumbnails change position on the illustration depending on their placement (left, right etc.)

Does that mean that, for example

- if `Left` is selected, you get:

image

if `Right` is selected, you get:

image

If so, would it be worth adding the assets in the design?
 
> Another question is if the illustrations change for light mode?

It does! :)

git-f0x:
Should the trackpad gestures dynamically update depending on workspace orientation and thumbnail placement (once those gestures are implemented/configurable in cosmic-comp)?
E.g. if workspaces are horizontal and on the bottom, a four-finger swipe up opens the workspace overview, but if they're on the top, a swipe down would open the overview (and similar for vertical workspaces).

acolombier:
> Should the trackpad gestures dynamically update depending on workspace orientation and thumbnail placement (once those gestures are implemented/configurable in cosmic-comp)? 

This is already the case. Are you suggesting it shouldn't?

git-f0x:
Oh, right, in the designs the text doesn't change, but in the video it works properly.
Though does it possibly make more sense to reverse the gesture direction (e.g. when workspaces are on the left, a swipe right opens them)?

acolombier:
> Assets are easy to add. 

Great. Please ping me when they are in so I can update the PR!

> Does the thumbnail placement actually change in the overview? If not, we should wait until it does before introducing the setting.

> Same thing with trackpad gestures. Do they actually change? 

Currently not - the setting is not applied yet. I don't quite know how it needs to be persisted for the `cosmic-comp` to pick it, assuming `cosmic-comp` already support it. 

Note I'm not developing live on COSMIC as well, so this work was mostly focused on the UI only. I might give it a go at some point using a VM (sadly I cannot run cosmic on my setup due to a [long standing issue](https://github.com/pop-os/cosmic-epoch/issues/135))

> The mockups you are looking at have been adjusted for first release with a lot of things removed from it.

Yes, most of the work I've done here is for the second release. AFAIK, Workspace as it stands is is already ready for the first release.

> About the bug in Iced - we should wait merging those illustrations until the bug is fixed.

Agreed - this is already in the PR description 

image

I have updated the PR title so it is more clear.

acolombier:
Since `libcosmic` is now based on iced 0.13, the SVG bug has been resolved.

I have also added the extra assets for right and bottom workspace overview thumbnails as @maria-komarova requested.

[Screencast from 07-11-24 21:41:55.webm](https://github.com/user-attachments/assets/4044e85c-7d00-47bd-988e-c90fccb625cc)

git-f0x:
Please remove or comment out non-functional settings, like the **Workspace Overview Thumbnails** section and Thumbnail placement. The App Library gesture also isn't implemented yet afaik, so that should also probably be commented out.
Also, the Trackpad gestures item should only show up if the system has a touchpad. See: https://github.com/pop-os/cosmic-settings/blob/42a3061da0354f0e461b95cfdb80bc2116fd80c7/cosmic-settings/src/pages/input/mod.rs#L204-L221

git-f0x:
You disabled the **Multi-monitor Behavior** section, instead of the **Workspace Overview Thumbnails** (`workspace_overview()`) section.
The **Workspaces Orientation** section will have to be refactored to only show the left and top workspace orientation images (depending on selection), and not use the `page.comp_workspace_config.workspace_thumbnail_placement` variable, since it doesn't exist in `cosmic-comp` and `cosmic-workspaces-epoch` don't support bottom and right positions yet.

acolombier:
Thanks for the quick review @git-f0x - I have commented out features unsupported by the cosmic-workspaces-epoch. I'm going to try to add the orientation in there, the multi-monitor behaviour is probably something I won't be able to do as I'm still struggling to understand the cosmic-comp code. Is that something you could give me pointers to or do you reckon it's best letting someone else look into it?

acolombier:
`Workspace Overview Thumbnails` is supported and functional in  cosmic-workspaces-epoch AFAICT - `Multi-monitor Behavior` isn't

> The Workspaces Orientation section will have to be refactored to only show the left and top workspace orientation

That should already be the case (I've hidden the dropdown for left/right, top/bottom selection)

git-f0x:
>`Workspace Overview Thumbnails` is supported and functional in cosmic-workspaces-epoch AFAICT

Those settings were there previously, but were removed because they were non-functional. Toggling that doesn't change anything in workspaces.

>`Multi-monitor Behavior` isn't

It's supported, and is present in the current master.

>That should already be the case (I've hidden the dropdown for left/right, top/bottom selection)

The images cannot show up, since the chosen image depends on a non-existing variable, and that variable has to be removed for Settings to compile.

acolombier:
> > `Workspace Overview Thumbnails` is supported and functional in cosmic-workspaces-epoch AFAICT
> 
> Those settings were there previously, but were removed because they were non-functional. Toggling either doesn't change anything in workspaces.

Weird - this is working for me. Changing horizontal/vertical does change where the workspace thumbnails get displayed

> > `Multi-monitor Behavior` isn't
> 
> It's supported, and is present in the current master.

Ah great. I will bring it back then.

> > That should already be the case (I've hidden the dropdown for left/right, top/bottom selection)
> 
> The images cannot show up, since the chosen image depends on a non-existing variable (`page.comp_workspace_config.workspace_thumbnail_placement`), and that variable has to be removed for Settings to compile.

This PR depends of [this one](https://github.com/pop-os/cosmic-comp/pull/811) which does add the relevant settings. I've just implemented the support for right/bottom placement on `cosmic-workspaces-epoch` and testing right now. Should submit a PR in the next few hours

git-f0x:
>Weird - this is working for me. Changing horizontal/vertical does change where the workspace thumbnails get displayed

I'm referring to the **Workspace Overview Thumbnails** section, with show name and show number, not the **Workspaces Orientation** section for choosing vertical/horizontal workspaces.

acolombier:
> with show name and show number, not the **Workspaces Orientation** section 

Ah sorry, it's late over there. I've noticed that `cosmic-workspaces-epoch` does display the workspace number under the thumbnail, so I'll see if I can at least implement number toggle.

image

acolombier:
I have implemented the features in [cosmic-workspaces-epoch](https://github.com/pop-os/cosmic-workspaces-epoch/pull/104) so I have restored the settings. I added the `cosmic-workspaces-epoch` as a dependency above, although the order release doesn't matter, only `cosmic-comp` setting definition needs to go first

PR #560, opened 2024-09-04
[Window management] Add Focus Navigation section

git-f0x:
This adds the **Focus follows cursor** and **Cursor follows focus** toggles.

### DIFF
--- a/cosmic-settings/src/pages/desktop/window_management.rs
+++ b/cosmic-settings/src/pages/desktop/window_management.rs
@@ -12,19 +12,43 @@ use cosmic_settings_page::Section;
 use cosmic_settings_page::{self as page, section};
 use slab::Slab;
 use slotmap::SlotMap;
+use tracing::error;
 
 #[derive(Copy, Clone, Debug)]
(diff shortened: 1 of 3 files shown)


mmstick:
Thanks!

GetPsyched:
Shouldn't one of these options be on by default?

git-f0x:
Neither option is the default behavior on any desktop environment, so they shouldn't be enabled by default.

GetPsyched:
That's correct. Though most tiling WMs have this on by default. Would it be possible to make this setting dependent on floating/tiling choice? Not automatically but using a dropdown with: "On if tiling", "On", "Off". Though I see how this gets too complex too quickly. Feel free to ignore my suggestion.

git-f0x:
I personally do find **Focus follows cursor** a bit weird when floating (and useful when tiling), so an option like that could be potentially useful.
But it would introduce more complexity, and the functionality in `cosmic-comp` might have to be changed to somehow be per-workspace (since tiling is per-workspace).

PR #564, opened 2024-09-06
Add Bluetooth screen

### LINKED ISSUE #437: Settings > Bluetooth
Settings page that allows to manage settings for Bluetooth devices. 
Bluetooth

[Figma file with more details](https://www.figma.com/design/Z6Z9Ox6WNkiiM8gcAkcElS/COSMIC-Settings-First-Release---Developer-Handoff?node-id=13857-1180&t=afTfg818R0XgQy4L-1).

acolombier:
https://github.com/user-attachments/assets/3495d248-06bb-48c3-b4cf-518b612432e7

(Running side by side with current bluetooth setting, for reference)

Closes #437 
Depends on https://github.com/pop-os/dbus-settings-bindings/pull/18
Depends on https://github.com/pop-os/libcosmic/pull/601/files

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -815,6 +815,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "bluez-zbus"
+version = "0.1.0"
+source = "git+https://github.com/acolombier/dbus-settings-bindings?branch=feat/bluez#71be7d32bb7a939c05f1ca8ecfbdf660b71ea139"
+dependencies = [
+ "futures",
+ "zbus 4.4.0",
+]
+
+[[package]]
 name = "borsh"
 version = "1.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1447,14 +1456,14 @@ dependencies = [
 [[package]]
 name = "cosmic-config"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/libcosmic#71cd25c06d230a742ebf660297478b732cf1882b"
+source = "git+https://github.com/acolombier/libcosmic?branch=feat/allow-toggler-to-be-disabled#51b7a946572493afd0ab5cdfbea8e7745e92e588"
 dependencies = [
  "atomicwrites",
  "cosmic-config-derive",
  "cosmic-settings-daemon",
  "dirs",
  "futures-util",
- "iced_futures",
+ "iced_futures 0.12.0 (git+https://github.com/acolombier/libcosmic?branch=feat/allow-toggler-to-be-disabled)",
  "known-folders",
  "notify",
  "once_cell",
@@ -1469,7 +1478,7 @@ dependencies = [
 [[package]]
 name = "cosmic-config-derive"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/libcosmic#71cd25c06d230a742ebf660297478b732cf1882b"
+source = "git+https://github.com/acolombier/libcosmic?branch=feat/allow-toggler-to-be-disabled#51b7a946572493afd0ab5cdfbea8e7745e92e588"
 dependencies = [
  "quote",
  "syn 1.0.109",
@@ -1553,6 +1562,7 @@ dependencies = [
  "as-result",
  "ashpd 0.9.1",
  "async-channel",
+ "bluez-zbus",
  "chrono",
  "clap",
  "color-eyre",
@@ -1651,7 +1661,7 @@ source = "git+https://github.com/pop-os/cosmic-settings-subscriptions#90df5c4b22
 dependencies = [
  "cosmic-dbus-networkmanager",
  "futures",
- "iced_futures",
+ "iced_futures 0.12.0 (git+https://github.com/pop-os/libcosmic)",
  "itertools 0.13.0",
  "libpulse-binding",
  "log",
@@ -1721,7 +1731,7 @@ dependencies = [
 [[package]]
 name = "cosmic-theme"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/libcosmic#71cd25c06d230a742ebf660297478b732cf1882b"
+source = "git+https://github.com/acolombier/libcosmic?branch=feat/allow-toggler-to-be-disabled#51b7a946572493afd0ab5cdfbea8e7745e92e588"
 dependencies = [
  "almost",
  "cosmic-config",
@@ -1837,7 +1847,7 @@ version = "0.19.0"
 source = "git+https://github.com/gfx-rs/wgpu?rev=20fda69#20fda698341efbdc870b8027d6d49f5bf3f36109"
 dependencies = [
  "bitflags 2.6.0",
- "libloading 0.8.5",
+ "libloading 0.7.4",
  "winapi",
 ]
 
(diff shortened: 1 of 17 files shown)


acolombier:
> Some questions to finish this PR
> 
> * Where to draw the line on what logic goes in `dbus-settings-bindings`?
> * Is there a loading spinner I can use instead of `...`?

Friendly ping @mmstick as you might have an idea for those two things :)

acolombier:
I realise the loading indicate need may not be obvious as the quality of the full screen is awful. Here is slightly better one

https://github.com/user-attachments/assets/2c5b9a6e-fbe2-4d8f-b245-7dc4bd5be836

cosmic-settings/src/pages/bluetooth/backend/dbus.rs
```diff
@@ -0,0 +1,322 @@
...
+use super::{Active, Adapter, Device};
+
+pub async fn get_devices<'a>() -> Result<Vec<Device>, zbus::Error> {
+    let connection = match Connection::system().await {
+        Ok(c) => c,
+        Err(e) => {
+            tracing::error!("zbus connection failed. {e}");
+            return Err(e);
+        }
+    };
```
mmstick, line 14: It would be good if each function accepts an already established connection. They're clonable so they can be freely passed around.

cosmic-settings/src/pages/bluetooth/backend/mod.rs
```diff
@@ -0,0 +1,5 @@
+use super::{Active, Adapter, Device};
+
+mod dbus;
```
mmstick, line 3: I think this whole module can go into the dbus-settings crate

mmstick:
There isn't a spinner widget yet. Either way, I don't think it'd be desired to hide the toggles. The connect buttons could be made insensitive with text that says "Connecting" though. Similar to what the WiFi page is doing at the moment.

acolombier:
> There isn't a spinner widget yet.

Ah that's a bummer! There is an example in libcosmic, but it looks quite outdated, so I couldn't make it work in reasonable amount of time.

> Either way, I don't think it'd be desired to hide the toggles.

Agreed, it was just to highlight the need tom provide feedback on loading. I have now added the ability to disable to a toggler in https://github.com/pop-os/libcosmic/pull/601/files

>  The connect buttons could be made insensitive with text that says "Connecting" though

I applied your suggestion, and I think it looks reasonable. WDYT?


https://github.com/user-attachments/assets/f180d55d-d656-45e8-a6ae-218af12ef9c1

acolombier:
This should now be ready.

The latest version should provide better code quality for the stream subscription. I took the freedom to add a setting popover to allow displaying devices without name.

acolombier:
Sadly, looks like an unrelated `iced` updated was merged in `libcosmic`, preventing building on `master`.

cosmic-settings/src/pages/bluetooth/backend/dbus.rs
acolombier: I have refactored the code and done so when possible. I've made the choice not to do so when moving in `tokio::spawn` for two reasons:
- Cleaner code - no need to make variable duplication + struct meber
- Re-instantiating the connection when needed will handle re-connection (e.g DBus crash)

cosmic-settings/src/pages/bluetooth/backend/mod.rs
acolombier: I moved as much as possible, but still kept a significant part, directly linked to local structs (`Adapter`, `Device`). Let me know if it is fine, happy to do more if you have specific suggestions.

maria-komarova:
Sorry, I can't see what the popover says that appears on clicking the cog icon. Would it be possible to see the screenshot?

acolombier:
Sorry, GH file limit is tough to work around for screencast.

image

image

maria-komarova:
No problem, thanks for the screenshots. Could you elaborate a bit on the rationale behind this feature for showing device without a name?

maria-komarova:
> An improvement on that would be to display the vendor name, inferred from the MAC address, (e.g An unknown Raspberry Pi Foundation device (28:CD:C1:xx:xx:xx)) to make it even more friendly (say you've just bought some Tuya Smartplug, they would show as An unknown Tuya Smart Inc. device (10:5A:17:xx:xx:xx)), but definitely out of scope for Alpha 1!

I like this suggestion is we can make it work reliably. 
In the meantime I don't think we should use a toggle. It seems to me we should show the devices with MAC address if this is the only identifier. Is there any other ID maybe? I wonder how is it handles in other systems... I assume people would still want to connect and they might not realize that they can't see the device in the list because it is missing a name. So the toggle seems like an unnecessary complication to me.

acolombier:
> I like this suggestion is we can make it work reliably.

The MAC address vendor assignation is public data AFAIK, so it would be fairly reliable. The risk is that some range will be registered under another name (e.g a BananaPI could use the `Broadcom` MAC)

> It seems to me we should show the devices with MAC address if this is the only identifier. 

Yes that is already the case. The default logic is to only show device that expose a friendly human identifier. The MAC address is only showed if the device has only broadcasted that information (and no friendly name) 

> Is there any other ID maybe?

Unfortunately no.

> I wonder how is it handles in other systems...

Usually they have some settings to show devices without name but this isn't the default (I remember seeing it somewhere on MacOS, but can't find with a Google search, so maybe it has been removed?). This is for Android:

image

> So the toggle seems like an unnecessary complication to me.

This would help keeping the list "clean" by default. In my case, if I enable the MAC address display, I can easily get 20+ devices after a few seconds of scan (I live in a flat, so basically detecting all the IoT devices of the neighbours :sweat_smile: ). Some devices are meant to be connected via companion app, and that would be why they don't provide a user friendly name. This means than connecting directly to it remains an edge case and would be great to be off by default. (like on Mac and Android)

maria-komarova:
It doesn't look like this setting is present on macOS or on iPhone. I can't find it anywhere there and web search confirms it. So they just don't show addresses with only MAC. 
I need to think more about this feature and where is the best place for it. I'll get back to you on that one.

mmstick:
Network settings pages have been merged now, so I'll help rebase this with changes that were made in that PR.

acolombier:
Thanks @mmstick :) 
I'll look at the rebase if it is okay as I have a significant amount of change locally which improve the quality of the code - I killed all the `tokio:spwan` in favour of `command::future` which allow to get operation result. 
Putting this to draft for now.

mmstick:
You should push the changes

acolombier:
I have rebased and refactor the code with a better approach, more stable. There is still some cleanup to do before merging. Note that the spacing seems to have changed, is that expected?

acolombier:
With this latest push, everything should be ready for review. 
Note that I haven't addressed the PIN prompt, which will require implementing a custom [Bluez Agent](https://github.com/pauloborges/bluez/blob/master/doc/agent-api.txt#L68), which is a significant amount of work, and thus should move to another PR. 
It may also be worth considering if there wouldn't be a better place to run the agent, rather the settings. My thought is, it would be great to be able to pair device directly for the bluetooth applet in the future, so perhaps a more central place to handle interaction (possibly file share in the future as well)

mmstick:
Rebased and merged earlier today via cbbbe9239397fe6e5c86ab26e9c935b5a41da596

mmstick:
Thanks!

Issue #472, opened 2024-08-09
[Feature Request] Add a "Disable" option for "Desktop > Window Management > Super Key"


nurho:
Based on searching questions on this issue for Gnome, many users (including myself) seem prefer the super key to behave purely as a modifier and not have an action on its own.

On Gnome there was a gsettings option to disable the overlay key but not a GUI option. I think the "Desktop > Window Management > Super key" setting would perfectly suit having a "Disable" option in the existing drop-down so that in addition to being able to change the super key behaviour, it can be disabled entirely.

Drakulix:
For what it's worth you can already do that in a config file.

Under `~/.config/cosmic/com.system76.CosmicSettings.Shortcuts/v1/custom` (which should exist if you modified any shortcuts), just add to the list of custom shortcuts:

```
    (
        modifiers: [
            Super,
        ],
    ): Disable,
```

I agree we should have a GUI option for this behaviour.

nurho:
Thanks for the tip @Drakulix! Good to know that there is a way but yes, a GUI option would still fit nicely here I think.

PR #573, opened 2024-09-08
Add "Disable" option for super key

jasonrhansen:
Implements #472


Drakulix (approved):

git-f0x:
Should the `super-key` text be changed to "Super key action", since it can seem like the Super key is disabled, rather than the action it performs on press?

jasonrhansen:
@git-f0x I agree that it's a bit confusing the way it is, and would be more clear if changed to "Super key action".

mmstick:
Thanks!

PR #577, opened 2024-09-09
[Appearance] Interface density

git-f0x:
This adds the 3 interface density options from the designs: Comfortable, Compact and Spacious.
Should be tested alongside pop-os/libcosmic#603.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,9 +54,7 @@ lto = "thin"
 cosmic-protocols = { git = "https://github.com/pop-os/cosmic-protocols//", rev = "bd65e0f2d55429954839c9e6c7bc5b0d198b85f0" }
 
 # For development and testing purposes
-[patch.'https://github.com/pop-os/libcosmic']
-libcosmic = { git = "https://github.com/git-f0x/libcosmic" }
-cosmic-config = { git = "https://github.com/git-f0x/libcosmic" }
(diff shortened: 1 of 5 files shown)


TheESN:
Can you show screenshots with examples of these settings please?

git-f0x:
This is pretty much ready, as far as the Settings side is concerned. Could possibly be merged with the `sections.insert(interface_density()),` line being commented out, or wait until everything in libcosmic and apps is polished.
Though merging as is would enable easier testing of what needs tweaks for all COSMIC components, but would introduce something that is a work in progress to users.

mmstick:
I think interface density should be part of Experimental Settings for now.

git-f0x:
I'll try to fix some App Library issues when Spacious, and then I'll try moving it to the Context drawer (above or below icon themes?).

mmstick:
I'll do some testing. Perhaps we don't need it in experimental settings if it's already working well.

git-f0x:
It looks slightly ugly in some cases, but when apps update their libcosmic version, there shouldn't be many things that are cut off or broken. I haven't found anything like that, other than in the App Library, and App Tray (my PR in applets fixes this).

mmstick:
It seems good enough to me.

mmstick:
Thanks!

git-f0x:
I'll work on fixing the App Library, so there isn't missing text on group buttons when Spacious.

PR #589, opened 2024-09-13
[WiFi] Adaptive WiFi icons

git-f0x:
Makes the WiFi icons in the list use the correct icons, that change depending on signal strength.


mmstick (approved):

mmstick:
Thanks!

PR #592, opened 2024-09-14
[Keyboard]: option to remap keys to caps lock

MaxLeoSchmid:
Adds option to remap different keys to caps lock like super, escape or backspace as well as swap caps lock and escape.


mmstick (approved):

mmstick:
Thanks!

PR #600, opened 2024-09-16
chore: reorder pages to match the design

GetPsyched:
This is based on the assumptions that these are accidental deviations and that the design is being followed strictly


GetPsyched:
Was fixed as a by-product of https://github.com/pop-os/cosmic-settings/commit/c6cd78ec9cf2a3a1522bfd724c66d81db5e1185a

PR #502, opened 2024-08-17
[draft] feat: show extended input sources (via toggle)

jmbuhr:
Needs https://github.com/pop-os/xkb-data/pull/4

Todo

- [ ] fix layout of toggle button so that its description doesn't use multiple lines
- [ ] localization of the description

PR #610, opened 2024-09-18
feat: support extended input sources

### LINKED ISSUE #606: [Feature Request] Support Extra Keyboard Layouts
A lot of people use keyboard layouts not listed in the X11 Base Rules, but rather the Base Rule Extensions.
An example would be the [EurKEY](https://eurkey.steffen.bruentjen.eu/) layout that I use. Currently this is only selectable through a config edit.

I'd love to see a toggle to include these extra keyboard layouts in the list of input sources.
Regardless of the toggle they should also be displayed in the active layouts section.

[xkb_data](https://docs.rs/xkb-data/latest/xkb_data/) already exposes these other layouts through the `extra_keyboard_layouts` function.

FlorianPallas:
Fixes #606, Alternative for #502


mmstick (approved):

mmstick:
Thanks!

PR #614, opened 2024-09-20
fix(page_list_item): match padding to design

git-f0x:
Adjusted to match a 100% scaled image from the designs.
Also removes the fixed height, which caused text to be cut off at smaller window widths.


mmstick (approved):

mmstick:
Thanks!

PR #630, opened 2024-09-25
[Networking] Match page list item to design

### LINKED ISSUE #627: Interface density settings in incorrect order
Usually density settings should either be smallest to largest, or largest to smallest

Right now it's middle, smallest, largest

screenshot-2024-09-25-04-15-01

git-f0x:
Adds an `info` argument to `page_list_item` that displays next to the arrow. `min_height` wasn't used for the container because it includes padding, so I added 8px vertical padding to a fixed size (20) component, to set it to 36 to match designs.
It would probably be better if padding around the icons could be set, since that would eliminate the need for a few `.padding` calls.
It has minor text wrapping issues, in that the text on the left doesn't wrap reliably if it doesn't have a description (not sure how to fix).
screenshot-2024-09-25-13-47-26
Should the "Connected", etc. text display only when there are multiple adapters of a type, rather than always?
Also includes various minor fixes.

Edit: Fixes #627 (order checked with Maria).

### DIFF
--- a/cosmic-settings/src/app.rs
+++ b/cosmic-settings/src/app.rs
@@ -931,6 +931,7 @@ impl SettingsApp {
                     page_list.push(crate::widget::page_list_item(
                         sub_page.title.as_str(),
                         sub_page.description.as_str(),
+                        "",
                         &sub_page.icon_name,
                         entity,
                     ))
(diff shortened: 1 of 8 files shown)


mmstick (approved):

mmstick:
Thanks!

Issue #664, opened 2024-10-04
[Help contributing] Adding entry in workspaces page


TornaxO7:
Hello!
I'd like to contribute to cosmic-settings by adding a spin button below where you can set the minimal amount of open workspaces:

![image](https://github.com/user-attachments/assets/bb1722a6-dead-4e44-a0c2-6ce469350821)

I think I found out, how to add a spin button in the frontend but I'd like to ask you guys first:

- Do you agree with this idea? My motivation is the following: In `i3` I used to have 6 open workspaces even if they don't have any windows in it. Some workspaces are there for a purpose so I navigate to the workspace *first* and then open my applications there. This isn't possible in cosmic because there are always at least 2 workspaces open.
- I'd create a category `Miscellaneous` under `Desktop > Workspaces` would you agree here?
- Now I'd need to find a way how to apply the value which is set in the settings to the cosmic expierence. Any tips on where I should look would be awesome :)

PR #665, opened 2024-10-04
Desktop > Workspaces: Add miscellaneous section

TornaxO7:
I couldn't find any guide of how to start contributing so I just started and tried it out.

This PR adds two more options to the `Desktop > Workspaces` page (look at the bottom):

image

- `Minimum open workspaces`: Users can set a minimal amount of workspaces which should be open during a session, even if they are empty. (Fixes https://github.com/pop-os/cosmic-settings/issues/664)
- Well, the second option already includes a description but is for the following use case:
    Assuming you have at least 3 workspaces: `1 [2] 3` where `[...]` symbolizes the current focussed workspace. If this option is set to `false` and you move to `workspace 1`, the content of workspace 3 doesn't move to workspace 2 automatically.

### DIFF
--- a/cosmic-settings/src/pages/desktop/workspaces.rs
+++ b/cosmic-settings/src/pages/desktop/workspaces.rs
@@ -6,7 +6,7 @@
 use cosmic::{
     cosmic_config::{self, ConfigGet, ConfigSet},
     iced::Length,
-    widget::{radio, settings, text},
+    widget::{radio, settings, spin_button, text},
     Apply, Element,
 };
 use cosmic_comp_config::workspace::{WorkspaceConfig, WorkspaceLayout, WorkspaceMode};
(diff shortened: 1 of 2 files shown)


mmstick:
Requires @pop-os/ux approval from @maria-komarova

TornaxO7:
> Requires @pop-os/ux approval from @maria-komarova 

It's offtopic but I'd like to ask:
- What's the best way to help you guys?
- is it more comfortable for you if there would be more PRs for already created issues?
- is it fine to just create a suggestion and open up a PR?
- Is there a recommended platform to communicate the development of cosmic, like where you can ask to help yan quickly ask the devs some questions or to see which tickets have which priority where the community can help?

maria-komarova:
I don't think we want to add this section to the settings. We have an idea for the feature we want to work through even though it is not an immediate priority. There are a lot of questions to cover when it comes to a set number of workspaces and we'd like to make sure we do a thorough job thinking through those. 

We're in alpha 3 focusing on the remaining features planned for the first release. So the best way to contribute is to find a repo you're interested in, then filter issues by alpha 3. Helping with milestone issues will get us to release faster.

TornaxO7:
Okeydokey!

PR #666, opened 2024-10-05
Add density to reset to default

git-f0x:
This is to reset the related CosmicTk configs to the defaults.
Also fixes the styling of the extended input sources toggle.
Not entirely sure about the second commit, but it seems to work/behave the same as roundness.


mmstick (approved):

mmstick:
Thanks!

PR #675, opened 2024-10-08
feat: Add alt-shift-tab reverse window switcher shortcut

skylerberg:
This is part of https://github.com/pop-os/cosmic-launcher/issues/173

I only added English due to not knowing how to translate into the rest of the languages.


git-f0x:
I think this can be closed, since the shortcut was added since.

PR #695, opened 2024-10-13
Add the ability to remap caps lock to control

green7ea:
The ability to remap the caps lock key was recently added but it wasn't possible to remap it to act as an extra control key. This commit makes that possible.

This is especially useful to reduce stress on the pinky finger when using control heavy applications like Emacs.


green7ea:
image

Generates the config

```(
    rules: "",
    model: "",
    layout: "us",
    variant: "colemak",
    options: Some(",caps:ctrl_modifier"),
    repeat_delay: 600,
    repeat_rate: 25,
)```

green7ea:
The settings app was tested but not its behaviour in the compositor as I'm still running alpha-2

mmstick (approved):

PR #706, opened 2024-10-18
Iced 0.13

wash2:


Cargo.toml
```diff
@@ -11,16 +11,18 @@ cosmic-randr = { git = "https://github.com/pop-os/cosmic-randr" }
 tokio = { version = "1.40.0", features = ["macros"] }
 
 [workspace.dependencies.libcosmic]
-git = "https://github.com/pop-os/libcosmic"
 features = [
```
jackpot51, line 14: If some of these features can be made optional but default, then it is likely cosmic-settings would run fine on any OS.

wash2: ok, i can probably do that now.

wash2: I added these as default, but they can be disabled "wayland", "dbus-config", "single-instance", "a11y", "wgpu"

mmstick (approved):

PR #710, opened 2024-10-19
fix(power): Fix battery state / charge  devices info.

### LINKED ISSUE #561: Power & Battery show that any connected devices are currently charging
Neither my mouse or headphones are currently charging. 

screenshot-2024-09-05-16-03-54

l-const:
Fixes #561.

Changes:

Before:

screenshot-2024-10-19-20-50-11

After:

screenshot-2024-10-19-20-15-40

Wrongfully using upower.on_battery (https://upower.freedesktop.org/docs/UPower.html) property instead of upower.device.state (https://upower.freedesktop.org/docs/Device.html) . The first one indicates whether it is a PC or laptop not if it is charging. The sibling properties of the object is LidIsClose :

screenshot-2024-10-19-21-01-34
screenshot-2024-10-19-21-01-55

I also found what KDE and lxqt is doing and what is using to derive this info:
1) https://github.com/lxqt/lxqt-powermanagement/issues/295#issuecomment-1015934354
2) https://github.com/KDE/solid/blob/master/src/solid/devices/frontend/battery.h#L201-L208 ( main KDE librarry abstration used for both)
3) An unrelated issue about the icon and this exact info into the main upower repo:  https://gitlab.freedesktop.org/upower/upower/-/merge_requests/20
4) Actual, implementation : https://gitlab.freedesktop.org/upower/upower/-/blob/master/src/linux/up-device-idevice.c?ref_type=heads#L285-294

Note: LQXT seems to have more complicated logic with the icons used, but nevertheless you can see what they use here to pick the right .svg: https://github.com/lxqt/lxqt-powermanagement/blob/master/src/iconproducer.cpp#L292-L300

```cpp
    switch (state)
    {
    case Solid::Battery::FullyCharged:
        svg.replace(QL1S("STATE_MARKER"), star);
        break;
    case Solid::Battery::Charging:
        svg.replace(QL1S("STATE_MARKER"), plus);
        break;
    case Solid::Battery::Discharging:
        svg.replace(QL1S("STATE_MARKER"), minus);
        break;
    default:
        svg.replace(QL1S("STATE_MARKER"), QString{});
    }
```

Sending the dbus message from terminal:
```bash
❯ dbus-send --print-reply             --system             --dest=org.freedesktop.UPower             /org/freedesktop/UPower/devices/battery_hidpp_battery_0     
        org.freedesktop.DBus.Properties.GetAll             string:org.freedesktop.UPower.Device | grep -C 2 "State"
      )
      dict entry(
         string "State"
         variant             uint32 2
      )

~
❯
```

2 -> Discharging

screenshot-2024-10-19-20-18-11
screenshot-2024-10-19-20-18-23

screenshot-2024-10-19-20-20-44

**Caveat:**

I only have a workstation , so I can't test the battery-applet (it doesn't show at all, but if there is a bug that there is it is hard to notice as i expect for laptop users to who only is visible to not show as charged and they might not notice that this is not changing when plugging it on power, or it might work because it is the main device?!!)  that I believe is also wrong here:

https://github.com/pop-os/cosmic-applets/blob/master/cosmic-applet-battery/src/app.rs#L104-L127

I did test the on_battery example of the upower_dbus crate where is showing for my workstation that is false -> that would be interpreted that my workstation /desktop PC was charging with the current code!!!:
screenshot-2024-10-19-21-13-27

Edit: Found gnome-shell's implementation here: https://github.com/GNOME/gnome-shell/blob/d60bdc95faf3607ec579558074cf1ca919db041a/js/ui/status/system.js#L19-L85

Like i said for the system battery (not peripherals) they use the DisplayDevice (an alias object for the system) and the Upower.device.state interface/property

```js
const BUS_NAME = 'org.freedesktop.UPower';
const OBJECT_PATH = '/org/freedesktop/UPower/devices/DisplayDevice';

const DisplayDeviceInterface = loadInterfaceXML('org.freedesktop.UPower.Device');
const PowerManagerProxy = Gio.DBusProxy.makeProxyWrapper(DisplayDeviceInterface);

................
...............


 // The icons
        let chargingState = this._proxy.State === UPower.DeviceState.CHARGING
            ? '-charging' : '';
        let fillLevel = 10 * Math.floor(this._proxy.Percentage / 10);
        const charged =
            this._proxy.State === UPower.DeviceState.FULLY_CHARGED ||
            (this._proxy.State === UPower.DeviceState.CHARGING && fillLevel === 100);
        const icon = charged
            ? 'battery-level-100-charged-symbolic'
            : `battery-level-${fillLevel}${chargingState}-symbolic`;

```

### DIFF
--- a/cosmic-settings/src/pages/power/backend/mod.rs
+++ b/cosmic-settings/src/pages/power/backend/mod.rs
@@ -1,7 +1,7 @@
 use chrono::{Duration, TimeDelta};
-use futures::future::join_all;
 use futures::FutureExt;
-use upower_dbus::{BatteryType, DeviceProxy};
+use futures::{future::join_all, TryFutureExt};
+use upower_dbus::{BatteryState, BatteryType, DeviceProxy};
 use zbus::Connection;
 
 mod ppdaemon;
@@ -230,7 +230,7 @@ pub struct Battery {
     pub icon_name: String,
     pub is_present: bool,
     pub percent: f64,
-    pub on_battery: bool,
+    pub is_charging: bool,
     pub remaining_duration: Duration,
 }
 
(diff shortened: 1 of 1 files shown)


mmstick (approved):

mmstick:
Thanks!

Issue #711, opened 2024-10-20
[BUG]: Input microphone slider doesn't do anything/work.


l-const:
Input microphone slider doesn't do anything/work and this is in contrast to the applet that does work and trigger cosmic-osd.

Demo:

[input-microphone-slider-settings.webm](https://github.com/user-attachments/assets/0fff573b-a9c8-47f8-9544-0dc528b61225)

l-const:
It appears that the **wpctl** (set-volume, mute) commands do not work for my particular **source**/**input** device even when I try to execute them from cosmic-term, the only command that works is this:

```bash
❯ wpctl status
PipeWire 'pipewire-0' [1.0.8, kostas@fe, cookie:928850085]
 └─ Clients:
        31. uresourced                          [1.0.8, kostas@fe, pid:27080]
        32. WirePlumber                         [1.0.8, kostas@fe, pid:27199]
        34. pipewire                            [1.0.8, kostas@fe, pid:27203]
        35. cosmic-osd                          [1.0.8, kostas@fe, pid:27089]
        38. xdg-desktop-portal                  [1.0.8, kostas@fe, pid:27350]
        46. MainConn                            [1.0.8, kostas@fe, pid:38720]
        47. WirePlumber [export]                [1.0.8, kostas@fe, pid:27199]
        63. cosmic-osd                          [1.0.8, kostas@fe, pid:54141]
        64. cosmic-settings                     [1.0.8, kostas@fe, pid:54141]
        67. cosmic-osd                          [1.0.8, kostas@fe, pid:38720]
        68. cosmic-osd                          [1.0.8, kostas@fe, pid:54141]
        70. MainConn                            [1.0.8, kostas@fe, pid:51829]
        72. cosmic-osd                          [1.0.8, kostas@fe, pid:51829]
        77. cosmic-osd                          [1.0.8, kostas@fe, pid:54141]
        78. wpctl                               [1.0.8, kostas@fe, pid:71267]

Audio
 ├─ Devices:
 │      48. TU106 High Definition Audio Controller [alsa]
 │      49. Built-in Audio                      [alsa]
 │
 ├─ Sinks:
 │      50. TU106 High Definition Audio Controller Digital Stereo (HDMI) [vol: 0.76]
 │  *   62. Built-in Audio Analog Stereo        [vol: 0.50]
 │
 ├─ Sources:
 │      73. Built-in Audio Analog Stereo        [vol: 0.74]
 │
 ├─ Filters:
 │
[...]

PR #713, opened 2024-10-20
fix(sound): Apply sound volume debounce.

### LINKED ISSUE #712: [BUG]: Sound volume de-bounce is not applied.
Sink and source variable is mixed up in some places and thus debounce is not applied,  look here:

Note: double-checked with the equivalent code in cosmic-applet-audio.

https://github.com/pop-os/cosmic-settings/blob/2a76db48e24a0de518a81945826da55f196e9a36/cosmic-settings/src/pages/sound.rs#L330-L345

l-const:
Fixes #712 . I was working on #711  and observed this, haven't figured out #711 yet though seems to be something different.

I double-checked with the almost same code of the cosmic-applet-audio here: https://github.com/pop-os/cosmic-applets/blob/master/cosmic-applet-audio/src/lib.rs#L381-L392

Edit: Just noticed in the applet we have a **64ms** delay whilst on settings **500ms**, which one of the two is correct?
There will be a significant noticeable difference with these changes applied between the two.
I think the 500ms feels kind of "laggy" when noticing cosmic-osd  whilst the 64ms seems like it is  applied almost "real-time" in my opinion. I could change the settings one in this PR if you have something in mind.

### DIFF
--- a/cosmic-settings/src/pages/sound.rs
+++ b/cosmic-settings/src/pages/sound.rs
@@ -316,7 +316,7 @@ impl Page {
             }
 
             Message::Pulse(pulse::Event::SourceVolume(volume)) => {
-                if self.sink_volume_debounce {
+                if self.source_volume_debounce {
                     return Command::none();
                 }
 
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks!

PR #725, opened 2024-10-27
Update cosmic_settings.ftl

bittin:
Fix typo

and will take care of updating the translation/the rest when i have more time and energy


leviport (approved):
Thanks for the fix!

bittin:
> Thanks for the fix!

no problem, will help continue the translation too when i have time, but i am busy rest of October, so won't be for Alpha 3, but will try to aim for Alpha 4 or 5 later in November-December

PR #728, opened 2024-10-27
[Panel] Fix opacity slider styling

git-f0x:
Makes the panel opacity slider show the current opacity, to match designs.
screenshot-2024-10-27-19-26-31

I fixed the spacing of rows with sliders, since spacing variables don't seem to be used there in the designs.
The sound page is a bit different, since there shouldn't be any spacing between the icon and number, but 8 spacing between the number and slider, so I put a `horizontal_space()` there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -230,9 +230,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.15"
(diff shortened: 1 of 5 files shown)


mmstick (approved):

mmstick:
Thanks!

PR #736, opened 2024-10-30
[Power] Only show suspend on battery power if the system has a battery

git-f0x:
This is a minor improvement to not show a non-applicable option on e.g. desktops.
I assumed that suspend on AC works the same on laptops and desktops, so I just added a separate fluent key to not mention "plugged in" on systems without a battery (since that's the only state they can be in).

Also fixed a minor text mismatch between the designs and implementation.


mmstick (approved):

mmstick:
Thanks!

PR #739, opened 2024-11-01
fix: condensed padding

### LINKED ISSUE #511: Containers overlap with scroll bar when window is narrowed to single-pane view
In the normal view, even when the width of the main pane is very narrow, there is a gap between the scroll bar and the right edge of the content containers:
screenshot-2024-08-20-19-57-26

When I squeeze the window down to a width where the left pane collapses, the scroll bar and the content containers touch each other:
screenshot-2024-08-20-19-57-14

```
$ apt policy cosmic-settings
cosmic-settings:
  Installed: 0.1.0~1724099986~22.04~6948373
  Candidate: 0.1.0~1724099986~22.04~6948373
  Version table:
 *** 0.1.0~1724099986~22.04~6948373 1002
       1002 http://apt.pop-os.org/staging/master jammy/main amd64 Packages
        100 /var/lib/dpkg/status
     0.1.0~1723893909~22.04~83a4296 1001
       1001 http://apt.pop-os.org/release jammy/main amd64 Packages
```

git-f0x:
This makes the min width horizontal padding "match" the designs with `@space_s`. It doesn't really match them, since all apps have 8 padding around them, so the previous value was technically more correct.
But since the padding around apps will likely be removed (so that the scrollbar is closer to the edge), it might be better to use the value from the designs (though the dynamic padding is more complex in the designs, depending on window width, etc...).
So this technically fixes #511, at least when not using Compact density (slight clipping with the scrollbar).

Initially wanted to try to fix [this issue](https://github.com/pop-os/cosmic-epoch/issues/1202), but wasn't able to, so I got carried away with the minor code cleanup. :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -318,7 +318,7 @@ checksum = "0ae92a5119aa49cdbcf6b9f893fe4e1d98b04ccbf82ee0584ad948a44a734dea"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.85",
+ "syn 2.0.86",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


mmstick (approved):

mmstick:
Thanks!

PR #764, opened 2024-11-06
feat: add Region & Language page

### LINKED ISSUE #12: Settings > Region and Language
Implement without "Advanced Formatting" for MVP. Limitations with how Linux stores formats requires revisiting advanced formatting possible options.

-- Carl

mmstick:
Closes #12

### DIFF
--- a/cosmic-settings/src/pages/desktop/appearance/mod.rs
+++ b/cosmic-settings/src/pages/desktop/appearance/mod.rs
@@ -408,8 +408,6 @@ impl Page {
         ]
         .spacing(theme.space_m())
         .width(Length::Fill)
-        .apply(widget::scrollable)
-        .height(Length::Fill)
         .apply(Element::from)
         .map(crate::pages::Message::Appearance)
     }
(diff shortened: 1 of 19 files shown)


git-f0x:
Might be a bit of an edge case, but I can't seem to set Serbian (Latin) as a language (it defaults to Cyrillic).
I think GNOME Settings on 22.04 provided such an option (maybe `lichen` doesn't support that yet?).

cosmic-settings/src/pages/desktop/appearance/mod.rs
```diff
@@ -408,6 +408,8 @@ impl Page {
         ]
         .spacing(theme.space_m())
         .width(Length::Fill)
+        .apply(widget::scrollable)
+        .height(Length::Fill)
```
git-f0x, line 412: These two additions aren't necessary after the recent context drawer changes, and lead to an infinite scrollbar.

git-f0x:
Sliders are also slightly broken for some reason. Not sure what libcosmic commit caused that.
screenshot-2024-11-11-14-58-17

mmstick:
Yeah I've been looking into a similar issue with FlexRow's child widgets having the same width as the FlexRow widget.

git-f0x:
The only somewhat notable change that could affect that between updates could be the Iced update (that includes this pop-os/iced@51ad918bf1b8ec1da0799217e917ae089a307d39 and the commit after it).

wash2 (approved):

PR #779, opened 2024-11-19
Updating cosmic-settings spin button widget

bhh32:
# Information
Creating this as a draft to prepare for the libcosmic Spin Button widget update.

PR #781, opened 2024-11-20
Update libcosmic

wash2:
some more updates on top of #779


git-f0x:
Would it maybe be better to use the `.title` method for the context drawer widget, rather than adding a `context_title` field?

PR #782, opened 2024-11-20
fix: increase channel size for display service

wash2:
cosmic-randr hangs if the channel fills up.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4009,9 +4009,9 @@ dependencies = [
 
 [[package]]
 name = "itoa"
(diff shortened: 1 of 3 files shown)


mmstick (approved):

mmstick:
Need to figure out why it blocks instead of awaiting later.

wash2:
> Need to figure out why it blocks instead of awaiting later.

I narrowed it down to block_on in a bunch of callbacks when calling send. https://github.com/pop-os/cosmic-randr/issues/9#issuecomment-2489518927

PR #860, opened 2025-01-04
Update Polish translation

mkljczk:


mmstick (changes requested):
There are conflicts with another translation PR that was merged for Polish. It may be worth using a diff tool to compare.

PR #862, opened 2025-01-06
feat: add fprint support

TitouanReal:
Adds basic fingerprint support:

Capture d’écran du 2025-01-06 09-32-00-obfuscated

I would like to finish this work and allow users to add and remove fingerprints. To continue, I would like official designs to follow.

Technical info useful for design:
- There can be multiple fprint readers.
- It is possible for zero or one of the readers to be marked as "Default".
- Adding a fprint requires its name (right-thumb, left-ring-finger, ...)

For more info, see [the doc](https://fprint.freedesktop.org/fprintd-dev/Device.html).


TitouanReal:
I have seen you put DBus bindinds in [here](https://github.com/pop-os/dbus-settings-bindings).  I believe you would prefer to have a fprint crate in there and use the crate in cosmic-settings.

mmstick:
Needs input from @maria-komarova then

TitouanReal:
The software stack for fingerptinrs on Linux is very not secure. I won't list all the flaws here , starting with spoofing, but one can read multiple articles about it.

Consequently, I think the Cosmic desktop has to decide whether fingerprint login is supported or not. If it is, It is probably a good idea to inform users about the security flaws of fingerprint login.

Issue #878, opened 2025-01-09
[BUG] Wrong translation of "Automatically hide dock" in Polish


oskarkulinski:
### Description
"Automatically hide dock" in Polish is translated the same as "Automatically hide panel" to "Automatyczne chowaj panel"

PR #879, opened 2025-01-09
Fix polish translation of automatically hide dock

oskarkulinski:
In "Automatically hide panel" and "automatically hide dock"  in Polish both "panel" and "dock" are translated as panel.
Described in issue 
[BUG] Wrong translation of "Automatically hide dock" in Polish #878


wash2 (approved):

mkljczk (approved):

mmstick (changes requested):
Conflict with previous Polish translation update

mmstick (approved):

PR #932, opened 2025-01-30
feat: Add terminal application selection to default apps page

MiraiDevv:
Implement terminal app detection and selection functionality:
- Scan desktop entries for terminal emulators
- Add support for selecting default terminal application
- Retrieve current default terminal using XDG mime associations
- Enable terminal category in dropdown selection

There are still some things that need to be addressed and implemented, such as:

- in the cosmic file, when you right-click on a folder and click on open in terminal, the respective selected terminal will open (I couldn't implement this, because I had no idea how to do it exactly)

- the code may be spaghetti or extremely long, because I'm still learning Rust, so any corrections to the code to shorten it or make it better will be appreciated

- opening files inside the terminal


mmstick (approved):

mmstick:
Thanks!

PR #936, opened 2025-01-30
refactor: Improve terminal binary path detection

MiraiDevv:
Replace alternative path detection with more robust binary path finding
Use which crate to locate terminal binaries
Extract binary path from desktop entry
Simplify symlink creation for x-terminal-emulator

It is important to mention that:

    I will still send the PR regarding this in the cosmic-files repository
    the functionality is not 100% complete, in developer mode, when we open cosmic-settings and cosmic-files and I change the terminal and try to click on open terminal in cosmic-files, cosmic files will open the terminal that was already selected, so if I am in cosmic-terminal when I open it and switch to warp and click on open in terminal it will open the old terminal and not warp (which was selected) so we still need to implement some function for cosmic-files to get the changes in real time from the cosmic-settings terminal change
    but if you close cosmic files and hit cargo run again, it opens the correct terminal (which would be warp).

continuation and correction of the PR #935

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1685,6 +1685,7 @@ dependencies = [
  "udev",
  "upower_dbus",
  "url",
+ "which",
  "xkb-data",
  "zbus 4.4.0",
  "zbus_polkit",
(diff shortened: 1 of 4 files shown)


mmstick:
It would be better to use a standard mechanism for this. x-terminal-emulator is specific to Debian/Ubuntu. There is currently a proposal for extending the xdg specs to add a default terminal spec. This is probably what we will do. https://github.com/Vladimir-csp/xdg-terminal-exec

PR #937, opened 2025-01-31
i18n(nl) dif pop-os master to feikedonia master 

feikedonia:
feikedonia master is 55 commits ahead of pop-os master;
I want to know why that is;
so I want to diff my repo against pop-os master


jacobgkau:
If you use `git commit --amend` before you push (or before using `--force` to push), you can modify a single commit instead of repeating the commit message dozens of times.

feikedonia:
I use github in the browser.

mmstick:
Then you need to reset the master branch and start over.

PR #948, opened 2025-02-05
added numlock on swap with capslock

99702:
Hey guys, 

I added additional `Swap with NumLock' for caps lock.

### DIFF
--- a/cosmic-settings/src/pages/input/keyboard/mod.rs
+++ b/cosmic-settings/src/pages/input/keyboard/mod.rs
@@ -43,6 +43,7 @@ static ALTERNATE_CHARACTER_OPTIONS: &[(&str, &str)] = &[
 static CAPS_LOCK_OPTIONS: &[(&str, &str)] = &[
     ("Escape", "caps:escape"),
     ("Swap with Escape", "caps:swapescape"),
+    ("Swap with NumLock", "caps:numlock"),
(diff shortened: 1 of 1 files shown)


leviport:
This seems like a very obscure customization that few users would want to make. What would be the reason for swapping numlock and caps lock?

99702:
Yes but In my case, when I'm on a Zoom call and remote sharing while coding in IntelliJ (with IdeaVim), pressing Escape triggers the Zoom pop-up for the other person instead of working in my IDE. To solve this, I remapped Num Lock to act as Escape in IntelliJ. However, Num Lock is far from the main typing area, so I swapped Caps Lock with Num Lock to make it more accessible. I couldn’t just remap Caps Lock to Escape because it requires two presses to switch back to normal mode, but using Num Lock allows me to do it in one press.

leviport:
That makes sense. However, I think this swap is still very niche, and I'm not sure it will benefit enough users to justify it being exposed as a default option.

In Gnome, more advanced xkb changes could be made in gnome-tweaks. There is a community [COSMIC Tweaks utility](https://github.com/cosmic-utils/tweaks) being worked on at the moment, though. Maybe that would be a better place to work on more advanced xkb configuration options? It would probably be good to open an issue on that repo to get the thoughts of the people currently working on it before building whole new pages, but I'm sure they'd love the help.

PR #949, opened 2025-02-06
Add window edge snapping setting

Koranir:
Relies on https://github.com/pop-os/cosmic-comp/pull/1189

~~Creates a new section in the "Window Managment" page called "Floating Layout", which has the "Window snap-to-edge threshold setting" that defaults to `0` (no edge snapping).~~

Adds a "Floating windows gravitate to nearby edges" toggle in the "Window Managment" page that sets the `edge_snap_threshold` cosmic-comp configuration to 10 pixels if enabled, or 0 pixels (off) if disabled.

### DIFF
--- a/cosmic-settings/src/pages/desktop/window_management.rs
+++ b/cosmic-settings/src/pages/desktop/window_management.rs
@@ -78,7 +78,7 @@ impl Default for Page {
             })
             .unwrap_or(true);
 
-        let window_snap_threshold = comp_config
+        let edge_snap_threshold = comp_config
             .get("window_snap_threshold")
             .inspect_err(|err| {
                 if !matches!(err, cosmic_config::Error::NoConfigDirectory) {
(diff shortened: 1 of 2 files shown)


Drakulix:
(@pop-os/ux please confirm, if I paraphrase this correctly.)

I think we just want this setting to be a toggle. It should display being "on" on any positive value (we want to allow people to mess with the config file) and "off" for '<= 0', toggling on should set the value to 10 and off to 0.

As a description we want "Display edge gravity for floating windows" instead of "Window snap-to-edge threshold". (@pop-os/ux Maybe the "for floating windows" part is redundant given this is in the "Floating Layout" section?)

Koranir:
I've taken the liberty of naming the setting "Floating windows gravitate to nearby edges", which I think represents what it does much better than "Display edge gravity for floating windows" - let me know if I have to change it back.

Drakulix (approved):
Code wise this looks good!

PR #952, opened 2025-02-07
improve(bluetooth): ignore devices with unknown type and without alias

abachmann:
The list of nearby bluetooth devices without an alias (only MAC address is displayed) is very long in my case (20+), making it difficult to select a device for pairing, especially since new devices are constantly being added and removed.

I installed blueman to see how it handles these devices, and it also does not display them.

Instead of ignoring all devices without an alias, I applied the filter only to unknown device types.

I also considered adding an option to enable/disable the filtering, but I'm not sure if it's really necessary or if anyone would actually use it.

### DIFF
--- a/cosmic-settings/src/pages/bluetooth/backend.rs
+++ b/cosmic-settings/src/pages/bluetooth/backend.rs
@@ -158,6 +158,8 @@ impl PartialEq for Adapter {
 
 impl Eq for Adapter {}
 
+const default_device_icon: &str = "bluetooth-symbolic";
+
 fn device_type_to_icon(device_type: &str) -> &'static str {
     match device_type {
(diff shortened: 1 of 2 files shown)


abachmann:
To ensure a consistent view of nearby devices, the bluetooth-applet also needs to be adjusted.

Are there any plans to have a shared library between cosmic-settings and the applets? I saw the [cosmic-settings-subscriptions](https://github.com/pop-os/cosmic-settings-subscriptions) repository.

mmstick:
That is being shared between cosmic-settings and applets. Not all applets have been refactored to make use of the shared logic though.

mmstick (approved):

mmstick:
Thanks!

PR #975, opened 2025-02-14
Respect default terminal in cosmic_settings_config (keybind SUPER+T) when changed

ellieplayswow:
**changes tl;dr**
- you can change your default terminal via COSMIC settings, but cosmic_settings_config is by default bound to `cosmic-term`
- when we change the default terminal, persist this setting if possible by grabbing the relevant `exec` from its FreeDesktop path

demonstration vid:

https://github.com/user-attachments/assets/2d7b0592-a440-422c-837f-7b4e75e5ed30

resolves https://github.com/pop-os/cosmic-epoch/issues/1557

### DIFF
--- a/cosmic-settings/src/pages/system/default_apps.rs
+++ b/cosmic-settings/src/pages/system/default_apps.rs
@@ -12,8 +12,13 @@ use cosmic::{
     widget::{self, dropdown, icon, settings},
     Apply, Element, Task,
 };
+use cosmic_config::{ConfigGet, ConfigSet};
+use cosmic_settings_config::shortcuts::action::System;
+use cosmic_settings_config::shortcuts::SystemActions;
 use cosmic_settings_page::{self as page, section, Section};
-use freedesktop_desktop_entry::{default_paths, DesktopEntry, Iter as DesktopEntryIter};
+use freedesktop_desktop_entry::{
+    default_paths, get_languages_from_env, DesktopEntry, Iter as DesktopEntryIter,
+};
 use mime_apps::App;
 use slotmap::SlotMap;
 use tokio::sync::mpsc;
@@ -81,6 +86,7 @@ pub struct AppMeta {
 pub struct Page {
     on_enter_handle: Option<cosmic::iced::task::Handle>,
     mime_apps: Option<CachedMimeApps>,
+    shortcuts_config: Option<cosmic_config::Config>,
 }
 
 impl page::AutoBind<crate::pages::Message> for Page {}
@@ -105,6 +111,10 @@ impl page::Page<crate::pages::Message> for Page {
     ) -> Task<crate::pages::Message> {
         if let Some(handle) = self.on_enter_handle.take() {
             handle.abort();
+        }
+
+        if self.shortcuts_config.is_none() {
+            self.shortcuts_config = cosmic_settings_config::shortcuts::context().ok();
         }
 
         let (task, on_enter_handle) = Task::future(async move {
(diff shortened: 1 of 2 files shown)


cosmic-settings/src/pages/system/default_apps.rs
```diff
@@ -239,6 +247,51 @@ impl Page {
                 if meta.selected != Some(id) {
                     meta.selected = Some(id);
                     let appid = &meta.app_ids[id];
+
+                    // if we're changing terminal, we want to update the COSMIC system actions shortcut
+                    // for terminal
+                    if category == Category::Terminal && self.shortcuts_config.is_some() {
+                        let shortcuts = self.shortcuts_config.as_ref().unwrap();
```
mmstick, line 254: Code must not unwrap

cosmic-settings/src/pages/system/default_apps.rs
```diff
@@ -239,6 +247,51 @@ impl Page {
...
+                                        let full_path = path.join([appid, ".desktop"].concat());
+                                        if full_path.exists() && full_path.is_file() {
+                                            resolved_path = Some(full_path);
+                                            break;
+                                        }
+                                    }
+
+                                    // if we find a valid .desktop file, we can grab its exec
+                                    if resolved_path.is_some() {
+                                        let desktop_entry = DesktopEntry::from_path(resolved_path.unwrap(), Some(&get_languages_from_env()));
```
mmstick, line 279: Another unwrap here

ellieplayswow:
@mmstick sorry about that, have resolved those

cosmic-settings/src/pages/system/default_apps.rs
```diff
@@ -239,6 +247,49 @@ impl Page {
...
+
+                    // if we're changing terminal, we want to update the COSMIC system actions shortcut
+                    // for terminal
+                    if category == Category::Terminal && self.shortcuts_config.is_some() {
+                        if let Some(shortcuts) = self.shortcuts_config.as_ref() {
+                            let system_actions = shortcuts.get::<SystemActions>("system_actions");
+                            match system_actions {
+                                Err(e) => {
+                                    // we don't have system_actions set?
+                                    println!("{}", e);
```
mmstick, line 259: This should be an error log:
```rs
tracing::error(?why, "unable to get system_actions shortcuts config");
```

cosmic-settings/src/pages/system/default_apps.rs
```diff
@@ -239,6 +247,49 @@ impl Page {
...
+
+                                    // if we find a valid .desktop file, we can grab its exec
+                                    if let Some(resolved_path) = resolved_path {
+                                        let desktop_entry = DesktopEntry::from_path(resolved_path, Some(&get_languages_from_env()));
+                                        if let Ok(desktop_entry) = desktop_entry {
+                                            if let Some(exec) = desktop_entry.exec() {
+                                                actions.insert(System::Terminal, String::from(exec));
+
+                                                if let Err(e) = shortcuts.set("system_actions", actions) {
+                                                    println!("{}", e);
```
mmstick: Also an error that should be a log instead of on stdout

ellieplayswow:
resolved all 3 - good shout re: canonicalisation

ellieplayswow:
sorting rustfmt

mmstick:
I need to check if this is writing only local modifications to `~/.config` instead of pulling values from the system.

ellieplayswow:
@mmstick no problem, let me know if i've missed anything here

mmstick:
Added a `get_local` method to the `cosmic_config::Config` so that we can get the local user config and only write what we've overwritten there.

mmstick (approved):

Issue #924, opened 2025-01-24
Feature: Additional scale options


maria-komarova:
![Image](https://github.com/user-attachments/assets/314e6356-a7d7-4092-95a9-66ae00c5cb2b)
[Figma mockups.](https://www.figma.com/design/Z6Z9Ox6WNkiiM8gcAkcElS/COSMIC-Settings-First-Release---Developer-Handoff?node-id=23165-4206&m=dev)

PR #976, opened 2025-02-14
implement additional scale options

uvera:
Designed like here https://github.com/pop-os/cosmic-settings/issues/924

image
image

This code isn't really ideal since I had to sync both slider and dropdown (and I'm still learning Rust :smile: ).

Also no gif/video since my OBS is acting up weird :shrug:


mmstick:
Figma designs show only 5 breakpoints. So there's too many here.

mmstick:
I'm wondering if the slider should only adjust the selected scale from the dropdown instead of the slider being the whole spectrum of values.

mmstick:
100, 150, 200, 250 would be less visually dense.

uvera:
50, 112, 175, 237.5 and 300 are the values if we go for even distribution with 5 breakpoints.

image

50, 100, 150, 200, 250, 300 are the values with 6 breakpoints even distribution

mmstick (approved):

PR #1000, opened 2025-02-22
i18n(de): add German translation

be-we:


i18n/de/cosmic_settings.ftl
```diff
@@ -127,6 +127,27 @@ bluetooth-available = Geräte in der Nähe
...
+            Super + = zum Hineinzoomen,
+            Super + - zum Herauszoomen,
+            Super + Scrollen mit deiner Maus
+    .increment = Zoom-Schrittweite
+    .signin = Lupe bei der Anmeldung starten
+    .applet = Lupe im Applet auf dem Panel ein-/ausschalten
+    .movement = Gezoomte Ansicht bewegt sich
+    .continuous = Kontinuierlich mit Zeiger
+    .onedge = Wenn der Zeiger die Kante erreicht
+    .centered = Zeiger zentriert halten
```
Drakulix, line 149: Would be nice to keep the wording here consistent with the translation in cosmic-comp:
https://github.com/pop-os/cosmic-comp/blob/master/resources/i18n/de/cosmic_comp.ftl#L1-L3

Also in the english version they are deliberately forming a sentence. "Gezoomte Ansicht bewegt sich Zeiger zentriert halten" isn't a valid german sentence.

be-we: How are translators supposed to know that there are other places with the same English text?

Drakulix: It is perfectly reasonable, that you can't just monitor every repo. That is why I am pointing it out.

be-we: That's all very well, but is there a translation memory for consistent wording?

Drakulix: No sorry. There isn't

be-we: In order to make the wording more consistent, I have created a translation memory. The term “translation memory” is a bit overstated, as it is only a Markdown table and far from complete.

If you are interested, you could perhaps take a look at it, make corrections, add wordings and mark the preferred translation, e.g., in bold:
https://github.com/be-we/cosmic-translation-memory/blob/master/de/cosmic_translation_memory.md

jackpot51: I can merge this once the discrepancy is resolved

PR #1001, opened 2025-02-23
improv: Make custom scale a slider & hide it unless asked for

### LINKED ISSUE #999: Additional scale options is confusing
As much as I love the extra granular scale options, I think there would be a better way to present it in the UI.

Image

Splitting the scaling into two like this makes it unclear whether the two percentages are summative or multiplicative. I know it is intended that 200% scale and 10% additional scale would mean a global scaling of 210%, but again, this is ambiguous. I initially assumed that it was taking 10% of 200% (20%) and then adding that onto the 200%, totalling to a 220% scale (sort of like a modifier for the previous option, not an addition to it).

Perhaps some sort of toast notification that reads "Scale set to 210%" would help clarify this? IMO, I feel like this is just overcomplicating the interface. A slider with a numerical spinner widget like KDE Plasma has would probably be more intuitive, while also allowing any value the user desires. Fractional scaling in COSMIC is pretty good from my experience, I don't think there would be a problem with allowing precise values instead of 5% increments. If single-digit percentage scaling isn't intended though, just a simple slider with steps would be okay.

Image

Koranir:
Fixes #999

https://github.com/user-attachments/assets/eb4cbc53-61d7-4037-88c8-888eedf8cd32

I have not tested this with multiple monitors yet, but it *should* work fine.

Due to how the page is loaded, if you set the custom slider to a value that matches a preset, it will reset to the preset value in the dropdown when the app is reloaded and hide the slider.

### DIFF
--- a/cosmic-settings/src/pages/display/mod.rs
+++ b/cosmic-settings/src/pages/display/mod.rs
@@ -960,6 +960,8 @@ impl Page {
 
     /// Set the scale preset of the active display.
     pub fn set_scale_preset(&mut self, option: usize) -> Task<app::Message> {
+        self.cache.scale_selected = Some(option);
+
         if scale_is_custom(option) {
             self.custom_scale = Some(self.config.scale);
 
@@ -974,7 +976,6 @@ impl Page {
             ScaleValue::Preset(value) => value,
         };
 
-        self.cache.scale_selected = Some(option);
         self.set_scale(scale)
     }
 
--- a/cosmic-settings/src/pages/display/mod.rs
+++ b/cosmic-settings/src/pages/display/mod.rs
@@ -700,7 +700,7 @@ impl Page {
                 ScaleValue::Preset(v) => format!("{v}%"),
             })
             .collect();
-        let selected_scale = DPI_SCALES
+        let mut selected_scale = DPI_SCALES
             .iter()
             .position(|s| matches!(s, ScaleValue::Preset(s) if *s == self.config.scale))
             .unwrap_or(0);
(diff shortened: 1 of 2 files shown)


l-const:
please make sure that the slider doesn't overflow when in really small dimensions , consider if you need to use flex_item anywhere like this commit: https://github.com/pop-os/cosmic-settings/commit/d1e41d53f65fc8add19d17af00bee601cd8404dc

Koranir:
> please make sure that the slider doesn't overflow when in really small dimensions , consider if you need to use flex_item anywhere like this commit: [d1e41d5](https://github.com/pop-os/cosmic-settings/commit/d1e41d53f65fc8add19d17af00bee601cd8404dc)

image

It *sorta* overflows at a maximal scale, but the scale can be changed by the dropdown too.

mmstick:
Changes to the UI require UX approval first

Bahasnyldz:
This needs an apply button.

WatchMkr:
Thanks for taking a look at this. A slider was initially planned but the slider widget does not feature stop points at different tick levels. We plan to add that in a later release and change the interface when it's available. The current slider can not be used because scaling in real-time to individual percentages is choppy and disorienting.

Koranir:
> A slider was initially planned but the slider widget does not feature stop points at different tick levels.

What does "different tick levels" mean?

The slider already can stop at certain increments, for example in my PR it steps by 5 percentage-point increments, unless the widget is completely broken in its implementation.

If the user wants to stop at a *25-percent* increment, they can just use the dropdown menu.

> The current slider can not be used because scaling in real-time to individual percentages is choppy and disorienting.

The scale is only applied once the drag is *released*, and dragging *only* changes the scale number next to the slider so the user can tell what value they are dragging to.

I could understand wanting an apply button, in which case I could happily add that.

PR #1025, opened 2025-03-01
feat(accessibility): replace magnifier zoom dropdown with slider

### LINKED ISSUE #1020: [FEATURE REQUEST] Use a slider for magnifier zoom instead of a list
Currently the magnifier zoom amount is controlled with a list of values, it might make more sense to switch this to a slider with increments of 5%

Image

HeitorAugustoLN:
Replace the list-based dropdown for magnifier zoom increment with a slider control that ranges from 25% to 200% with 5% increments. This provides more intuitive and flexible control over the zoom level.

It is my first time contributing here!

Closes #1020

image

### DIFF
--- a/cosmic-settings/src/pages/accessibility/magnifier.rs
+++ b/cosmic-settings/src/pages/accessibility/magnifier.rs
@@ -24,8 +24,7 @@ pub struct Page {
 
     accessibility_config: cosmic_config::Config,
     zoom_config: ZoomConfig,
-    increment_values: Vec<String>,
-    increment_idx: Option<usize>,
+    zoom_increment: u32,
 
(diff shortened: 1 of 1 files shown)


mmstick:
UI changes require UX approval first @pop-os/ux

maria-komarova:
Thanks for the contribution. I'm sorry but I don't think using this slider will make for a better experience here. For one thing, we don't yet have a slider with markers in libcosmic and we'll need one to show that the slider is not continuous and moves in increments. Another thing is the range you mentioned will result in too many increments for the slider. And we don't yet have enough evidence that 5% increments will better fulfill people's needs.

PR #1062, opened 2025-03-17
a11y: screen filters

Drakulix:
~~Draft while https://github.com/pop-os/cosmic-protocols/pull/56 isn't merged.~~

*Should* be merged after cosmic-comp support, but support in the compositor is optional, the UI can adopt.

### DIFF
--- a/cosmic-settings/src/pages/accessibility/mod.rs
+++ b/cosmic-settings/src/pages/accessibility/mod.rs
@@ -1,32 +1,61 @@
 use cosmic::{
+    Apply, Task,
     cosmic_theme::{CosmicPalette, ThemeBuilder},
     iced_core::text::Wrapping,
     theme::{self, CosmicTheme},
-    widget::{button, container, horizontal_space, icon, settings, text},
-    Apply, Task,
+    widget::{button, container, dropdown, horizontal_space, icon, settings, text, toggler},
 };
 pub use cosmic_comp_config::ZoomMovement;
 use cosmic_config::CosmicConfigEntry;
 use cosmic_settings_page::{
-    self as page,
+    self as page, Insert,
     section::{self, Section},
-    Insert,
 };
+use num_traits::FromPrimitive;
 use slotmap::SlotMap;
 
 pub mod magnifier;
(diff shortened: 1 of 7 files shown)


cosmic-settings/src/pages/accessibility/mod.rs
```diff
@@ -170,18 +203,74 @@ pub fn vision() -> section::Section<crate::pages::Message> {
...
+                })
+                .add({
+                    let selections = if page.screen_filter_selection == ColorFilter::Unknown {
+                        &page.screen_filter_selections
+                    } else {
+                        &page.screen_filter_selections[0..4]
+                    };
+                    cosmic::Element::from(
+                        settings::item::builder(&descriptions[color_filter_type]).control(
+                            dropdown(
```
ids1024, line 238: Should the dropdown be insensitive if the color filter toggle is off, or is intended to be able to select a filter, then enable it?

It should be insensitive when the protocol isn't present, anyway, like the toggles.

Drakulix: I didn't see a way in the api to make it insensitive. Though I think it is nice from an UX perspective to be able to set and enable it in any order.

ids1024: Hm, yeah. Looking at the implementing of `Dropdown` in libcosmic I don't see any way to make it insensitive. Definitely something we'll want in libcosmic, but not a blocker here.

cosmic-settings/src/pages/accessibility/mod.rs
```diff
@@ -170,18 +203,74 @@ pub fn vision() -> section::Section<crate::pages::Message> {
...
+                                selections,
+                                Some(page.screen_filter_selection as usize),
+                                move |idx| {
+                                    let filter =
+                                        unsafe { std::mem::transmute::<usize, ColorFilter>(idx) };
```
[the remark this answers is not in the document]
Drakulix: fixed

mmstick (approved):

PR #1072, opened 2025-03-21
Add Applications page

git-f0x:
This adds the Applications page in the NavBar, and moves default apps, startup apps and xwayland scaling to this page.
The placement in the NavBar is arbitrary, since I wasn't able to find where it should be in the designs.
I've also set `system-reboot-symbolic` as a placeholder icon for startup apps.


mmstick (approved):

mmstick:
Thanks!

Issue #1066, opened 2025-03-19
[translation bug] can't translate the magnifier's zoom increment "50% (default)" zoom level


feikedonia:
_view of the magnifier settings page_
![Image](https://github.com/user-attachments/assets/f797b84d-6736-4ec0-b09a-63faa8451e0c)

this minor issue resembles within the magnifier's zoom increment selector. 

_settings > accessibility > magnifier > zoom increment_
![Image](https://github.com/user-attachments/assets/f8973af2-d3e5-41b6-9320-b21a1c50c04b)

I can't translate the "(default)" string to Dutch "(standaard)".

*cosmic-settings i18n nl cosmic_settings.ftl*
![Image](https://github.com/user-attachments/assets/3676dcde-b8db-4ff1-8f0d-cc34a8669eec)

cappsyco:
Localisation support for this string merged here https://github.com/pop-os/cosmic-settings/pull/1076

PR #1076, opened 2025-03-23
1066

cappsyco:
Added a translation string for "Default", addressing https://github.com/pop-os/cosmic-settings/issues/1066

### DIFF
--- a/cosmic-settings/src/pages/accessibility/magnifier.rs
+++ b/cosmic-settings/src/pages/accessibility/magnifier.rs
@@ -1,23 +1,24 @@
 use std::collections::HashSet;
 
 use cosmic::{
-    Apply,
     iced::{Element, Length},
     iced_core::text::Wrapping,
     surface,
     widget::{self, icon, settings, svg, text},
+    Apply,
 };
 use cosmic_comp_config::{ZoomConfig, ZoomMovement};
 use cosmic_config::{ConfigGet, ConfigSet};
-use cosmic_settings_config::{Action, Binding, shortcuts};
+use cosmic_settings_config::{shortcuts, Action, Binding};
 use cosmic_settings_page::{
-    self as page, Entity,
+    self as page,
     section::{self, Section},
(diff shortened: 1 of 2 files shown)


cosmic-settings/src/pages/accessibility/magnifier.rs
```diff
@@ -69,9 +69,9 @@ impl Default for Page {
                     "{}%{}",
                     val,
                     if val == ZoomConfig::default().increment {
-                        " (Default)"
+                        format!(" ({})", fl!("default")).to_string()
```
mmstick, line 72: Since this is already within a formatter, this should be used instead https://doc.rust-lang.org/stable/std/macro.format_args.html

cosmic-settings/src/pages/accessibility/magnifier.rs
```diff
@@ -69,7 +68,7 @@ impl Default for Page {
                     "{}%{}",
                     val,
                     if val == ZoomConfig::default().increment {
-                        format!(" ({})", fl!("default")).to_string()
+                        format_args!(" ({})", fl!("default")).to_string()
```
mmstick, line 71: No need to use to_string with format args

cappsyco: Ah, I left that in as the blank string in the else complains otherwise but there's clearly a better way to do this. Sorry, made a bit of a mess with this PR!

mmstick: `format_args()!` gives you a `fmt::Arguments` value that will be lazily-evaluated by a formatter, so you can avoid the double allocation. Alternatively, you can conditionally construct strings with https://doc.rust-lang.org/stable/std/fmt/fn.write.html

cappsyco: Thank you!

mmstick (approved):

PR #1080, opened 2025-03-25
[Accessibility] Use popup_dropdown for color filters

git-f0x:
This makes the color filter dropdown able to go outside the window. Also fixes some padding issues, and changes the `go_next_item` and `go_next_with_item` functions to use `on_press_maybe` (Message argument changed to `impl Into<Option<Msg>>`), so that it could be used in the Accessibility page for the Magnifier.
Also renames the `legacy-applications` fluent key to `legacy-app-scaling` in non-updated translations, to avoid breakage and make it easier for translators (Hungarian excluded since there's already a PR).

The formatting was caused by switching to use `theme::spacing()`, but should have probably been in another commit (sorry!). This should format it to the 2024 edition, similar to libcosmic.


mmstick:
May need to do the spacing changes in a separate PR. Or to save without formatting changes.

git-f0x:
I've restricted the formatting and spacing changes to the files I touched. Also excluded pt-BR, since there's a PR for that now as well.

mmstick (approved):

mmstick:
I'll make formatting changes later once I merge shortcuts UI improvements

PR #1110, opened 2025-04-10
change `default_folder` path of system backgrounds based on distribution

thefossguy:
NixOS is not a Linux distribution that is compliant with the "File Hierarchy System." The top-level directory `/usr` does not exist on NixOS. So on NixOS, we won't find any backgrounds in the `/usr/share/backgrounds/` directory.

Instead, what one usually finds in `/usr/share` on FHS-compliant distributions, on NixOS, it is found in the
`/run/current-system/sw/share` directory.

So, check if we are on NixOS or a NixOS-like distribution by checking the `/etc/os-release` file. If we are on such distribution, use the `/run/current-system/sw/share/backgrounds/` directory. Otherwise, use the `/usr/share/backgrounds/` directory for getting system backgrounds.


alyssais:
A more portable solution here would be to look up using `XDG_DATA_DIRS`.

Issue #1128, opened 2025-04-16
[BUG] Editing user 'Full name' or 'Username' does not save


cappsyco:
The inline edits for 'Full name' and 'Username' on a user's edit / details screen do not save.

To replicate:

**Full name**
1. Inline edit the Full Name. It should update in the field and for the user on screen.
2. Switch to another tab and then back and the old name is reverted in the label and field
3. Check /etc/passwd to see the old name is still present

**Username**
1. Inline edit the username. It should update in the inline edit field
2. Switch to another tab and then back and the old username is reverted in the field
3. Check /etc/passwd (or other user files) and the old username is still present
4. Check the user's home directory and it is still labelled after the old username (although changing this might be a whole different consideration?...)

![Image](https://github.com/user-attachments/assets/92bc4eca-d281-4cac-b470-5ede993b172c)

cappsyco:
Sorry, closing this as it is not a bug, I've realised that you need to press 'Enter' to save changes.

WatchMkr:
I'd like to keep this open. A couple things are off.

1. Pressing Enter should not close the Users page.
2. Pressing Enter should not be necessary (clicking out of the field should be enough)
3. It appears the change is saved when you click out of the field but it's not

cappsyco:
As an update to this, points 2 and 3 have been fixed in the PR above. 

Point 1 is a wider issue that occurs across the Settings app and not unique to the Users section, but so far I've not been able to track down why this happens.

Issue #1127, opened 2025-04-16
Improve Users panel password change


WatchMkr:
Changing the password should require double entry to confirm the user entered the intended password.

https://www.figma.com/design/Z6Z9Ox6WNkiiM8gcAkcElS/COSMIC-Settings-First-Release---Developer-Handoff?node-id=13918-2014&t=lz5UNeWIzS7J2Jy8-1

cappsyco:
New password change flow implemented and merged: https://github.com/pop-os/cosmic-settings/pull/1129

Issue #844, opened 2024-12-27
Password change in settings very broken


oscardssmith:
Changing your password from the system and accounts menu seems deeply broken. Specifically, it appears to not actually change your password (and sometimes it seems like it might change your password to the wrong password).

bhh32:
Can you give more details, specifically abut the changing it to the wrong password? Where there specific special characters or anything else that it may have interpreted incorrectly?

oscardssmith:
I don't think so. it was all printable ASCII. I'm not entirely sure what happened since I changed it, and the time I realized that something was wrong was after a restart at which point neither the old nor new password seemed to work (although it is definitely possible that I mistyped the old or new password repeatedly)

novinxy:
I had similar experience with changing password. I wrote typo when I changed password and had to do recovery.
There are missing bits in UI/UX like hovers, etc. which might create confusion, here is list of potential issues I see:
- no hover on edit icons & hover on textbox. It might create confusion _"Should I click edit icon or just click on textbox ?"_
- no border for textbox. hard to  tell where input should be
- accepting changes work differently for 'Full name'/'Username' vs Password
   - for Full name & Username you can either press return key or click outside textbox to accept changes. Non technical people might not figure it out, in many cases people just prefer to have 'accept/ok' button. It might be nice addition.
   - password only works by pressing return key, clicking outside textbox doesn't change password. It works in different way and there is no information that change was applied or not. Only one method (hidden one) to accept is insufficient. Some people might know what to do
- to fix issue with writing 'typos' there are 2 simple additions that would help [See examples](https://www.google.com/search?client=firefox-b-d&sca_esv=2074af11bd466d63&q=change+password+dialog&udm=2&fbs=ABzOT_CWdhQLP1FcmU5B0fn3xuWpA-dk4wpBWOGsoR7DG5zJBr1qLlHFB6ZBcx-Arq68_wfw0s-Sy4efUF8x4O2idyUNU17MxNnWP-gPg-bH3QglMxJINyM4i32GPOhiyHOinjE1SNJ-lrwwfWI8tJPY6E6U6Q-vb3O-euuSKgF-383g6tbUXVM&sa=X&ved=2ahUKEwjswf_czcaLAxU5CBAIHd3uAkAQtKgLegQIGBAB&biw=1920&bih=962)
   - 'confirm password' textbox in which we make sure that user wrote password correctly
   - button to 'unhide' password. Without it, it's impossible to review password

[...]

PR #1129, opened 2025-04-16
feat: improved password workflow

cappsyco:
Implementing the following from https://github.com/pop-os/cosmic-settings/issues/1127

- User creation and password update now requires double entry of matching passwords
- New flow for changing password from inline edit to separate dialog
- Password fields converted to secure_input with visibility toggles as per the greeter and network applet.

There is a separate issue of passwords never being set or updated correctly. This PR is no more or less broken than it is currently as it uses the same password update function https://github.com/pop-os/cosmic-settings/issues/844

Another existing issue with user settings that's not currently addressed in this PR is the updating of full name and username https://github.com/pop-os/cosmic-settings/issues/1128

### DIFF
--- a/cosmic-settings/src/pages/system/users/mod.rs
+++ b/cosmic-settings/src/pages/system/users/mod.rs
@@ -32,8 +32,9 @@ pub struct User {
     id: u64,
     profile_icon: Option<icon::Handle>,
     full_name: String,
-    password: String,
     username: String,
+    password: String,
+    password_confirm: String,
     full_name_edit: bool,
     password_edit: bool,
     username_edit: bool,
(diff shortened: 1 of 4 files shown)


cappsyco:
Added a fix to the password saving (on new user creation and updating) which addresses the previous issue of the plain text version being saved in /etc/shadow and therefore making login for that user impossible. I've used this crate which I hope is suitable but I can change this if needed: https://crates.io/crates/pwhash

mmstick (approved):

mmstick:
Thanks!

mmstick:
Doesn't seem to work on Pop!_OS 24.04. Passwords are invalid.

cosmic-settings/src/pages/system/users/mod.rs
```diff
@@ -556,7 +557,7 @@ impl Page {
                 self.dialog = None;
 
                 let uid = user.id;
-                let password = user.password;
```
mmstick: Seems we need to check `/etc/login.defs` for the encryption method used, and then use that.

cappsyco: Solution for this https://github.com/pop-os/cosmic-settings/pull/1138

More pertinently, I have a better Pop_OS testing setup now :/

PR #1135, opened 2025-04-19
fix: inline editing behaviour

cappsyco:
Relating to issue https://github.com/pop-os/cosmic-settings/issues/1128

**Done**
- on_unfocus save event for full name and username, in addition to pressing enter

**Ongoing**
- Pressing enter will often (but not always) trigger a page change to the parent. This is something that's happening on any second level page, and in any state, across the Settings app so I'm currently unsure whether this originates in this repo or maybe somewhere in libcosmic. Looking into this more.


cappsyco:
If it's ok I would like to submit this for review as it is. The bug with the Enter key moving back a page is all through the Settings app and so not specific to this area and I think should be looked at separately.

wash2 (approved):

PR #1138, opened 2025-04-21
fix: use appropriate hashing method for passwords

cappsyco:
This ensures the appropriate hashing method is used based on the /etc/login.defs file. This is tested and working in Pop and Arch for user creation and password updating.

This is a followup to the discussion here: https://github.com/pop-os/cosmic-settings/pull/1129#discussion_r2052474934


mmstick (approved):

leviport (approved):
Seems to be working nicely on sha512

Issue #1120, opened 2025-04-14
Use XDG_DATA_DIRS to determine wallpaper path


Pandapip1:
https://github.com/pop-os/cosmic-settings/commit/2e2898b31fabcb3f3b561e6c4cea2aca9de9b284 adds a Nix OS-specific patch. But this is kind of ugly and adds a small but nonzero maintenance burden that most COSMIC developers aren't going to be testing for. It would be better to instead iterate through `XDG_DATA_DIRS`, which will work on all distributions, even ones with unconventional root directory structures that aren't NixOS (e.g. guix).

thefossguy:
This is possible but I didn't do it to cause any non-minor refactors being unfamiliar with the codebase.

PR #1159, opened 2025-04-30
iterate XDG_DATA_DIRS when loading backgrounds

ellieplayswow:
**changes tl;dr**
- instead of using `/usr/share/backgrounds` everywhere, instead (try to) use the first `$XDG_DATA_DIRS` that has a `backgrounds` subdirectory
- if none, fallback to `/usr/share/backgrounds`

functionality remains unchanged on pop 24.04


ellieplayswow:
fixes #1120

mmstick (approved):

PR #1161, opened 2025-05-01
i18n(zh-CN): update translation

pugaizai:
finally


pugaizai:
also https://github.com/pop-os/cosmic-comp/pull/1387

mmstick (approved):

PR #1218, opened 2025-05-28
i18n(nl) adding Dutch translations, but also modifying some punctuation to EN and PL.

feikedonia:


resources/com.system76.CosmicSettings.Mouse.desktop
```diff
@@ -4,11 +4,13 @@ Name[zh_CN]=鼠标
 Name[pl]=Mysz
 Name[hu]=Egér
 Name[pt]=Mouse
+Name[nl]=Muis
 Comment=Mouse speed, acceleration, and natural scrolling.
 Comment[zh_CN]=鼠标速度、加速度和自然滚动
 Comment[pl]=Prędkość myszy, przyspieszenie i naturalne przewijanie.
 Comment[hu]=Egérsebesség, gyorsítás és természetes görgetés.
 Comment[pt]=Velocidade do mouse, aceleração e rolagem natural.
+Comment[nl]=Muissnelheid, -versnelling en 'natuurlijk' scrollen.
```
feikedonia, line 13: 'natuurlijk' is in quotation marks because a reader may otherwise read it as 
"Mouse speed and acceleration, and scrolling of course!"
( of course and natural have the same translation in Dutch )

resources/com.system76.CosmicSettings.Input.desktop
```diff
@@ -4,8 +4,10 @@ Name[zh_CN]=输入设备
 Name[pl]=Urządzenia wprowadzania danych
 Name[hu]=Beviteli eszközök
 Name[pt]=Dispositivos de Entrada
-Comment=
+Name[nl]=Invoerapparaten
+Comment=Keyboard, cursor, etc.
```
feikedonia, line 8: Adds a English comment

phoekjelle:
When I think the English translation is also valid in Dutch, I wont add an extra entry: I assume that the English default is automatically displayed to the user.

@mmstick said: "No point adding if it is the same as the default key/value pair."

wash2 (approved):

phoekjelle:
@wash2  did you bundel all commits into one packet? 

e810acd8f15e3267960ebba1595427f87172ab0d

afbeelding
I can just Sync Fork this?

git-f0x:
You should try to minimize the amount of commits (by making a commit only when you're done with a logical whole), since 44 is a bit too much for just a localization update for one language (but they can be squashed, so it's not too much of an issue).
GitHub Desktop is quite simple to use, so you can just edit the files on your machine locally (since multiple files can't be edited at the same time on the site), and when you're done, make a commit in the app. But it isn't needed if you don't want to do it. It's just a bit cleaner (since something like Weblate isn't a thing yet).

You can click Sync fork and then discard your 44 commits, which should make your fork match upstream.

phoekjelle:
so, I could just use `git clone <url> `
What do I need to do to configure Git on my local machine?


So, I sould make some changes, then `git add <filename>` and when I'm ready, `git commit -m <message>` and at the end `git push` 

and in the github website I can then tell it to make a github "pull request" (PRs are a github only thing, right)

also, can I force cosmic settings, for example, to use the translation files I create, overwriting the official file downloaded with my distribution?

therealmate:
Alright, so a basic Git tutorial for Github:
Go to Github Settings>Developer Settings>Personal Access Tokens>Tokens (classic)>Generate new token>Generate new token (classic), add a name, change expiration date if you want, also add ```repo``` for scopes>Generate token>copy the token (used in later steps)
```git clone <your fork>```
Edit the files/translations
```git add .``` (adds everything, so no need to add files one by one)
```git commit -m "some commit message"```
```git push``` (If you do a rebase/squash multiple commits into one like in the commit above, you will need --force, but we are only focusing on basic commits for now)
Enter your username, for the password, paste the token that we created in the first step, and now, the fork should be updated on Github and is ready for creating the PR.
Let me know if you don't understand something

"also, can I force cosmic settings, for example, to use the translation files I create, overwriting the official file downloaded with my distribution?"
I don't think this is possible without a recompilation since the translations are hard-coded into the final binary

phoekjelle:
cool. thanks
Which app do you use to edit?

therealmate:
You can use any text editor to edit the files. Personally, I use Kate because I'm on Plasma, but if you're on COSMIC, then the COSMIC Text Editor might be a better choice.

PR #1230, opened 2025-06-07
i18n: fix duplicates and curly brackets

Lampan-git:
Fix various duplicates and fix curly brackets errors


mmstick (approved):

therealmate:
Actually, the Hungarian (hu) translation has a mistake now that I look at it closely, the removed line is not a duplicate, but a typo, it is supposed to be window-management-appearance instead of duplicating window-management

Lampan-git:
Looks like it's the same for italian and portuguese as well. I can open a new pr to fix this

PR #1232, opened 2025-06-08
feat: add simple bios version section to system/firmware page

nurseiit:
> disclaimer: I am a first time contributor to this project but a long time user who recently jumped to test out the 24.04 Alpha.

before: an empty page without any sections under `"System & Accounts" / "Firmware"` page

Screenshot_2025-06-08_16-25-00

after: added a single section with `"BIOS version"`

Screenshot_2025-06-08_16-28-11

---

NOTES:

- Gnome settings have it called "Firmware version" -- let me know if you prefer it renamed to that instead of "BIOS Version"
- The other thing missing from this page is "Kernel version", which I can also add but wanted to verify if everything is OK with this PR first
- One thing I don't like about this is that I reused the `Info` from `About` page which loads up a bunch of other stuff too -- if you'd prefer to have it separated to a different place then please do let me know

Thanks!

### DIFF
--- a/cosmic-settings/src/app.rs
+++ b/cosmic-settings/src/app.rs
@@ -646,6 +646,12 @@ impl cosmic::Application for SettingsApp {
                         return page.update(message).map(Into::into);
                     }
                 }
+
+                crate::pages::Message::Firmware(message) => {
+                    if let Some(page) = self.pages.page_mut::<system::firmware::Page>() {
+                        return page.update(message).map(Into::into);
+                    }
+                }
             },
 
             #[cfg(feature = "wayland")]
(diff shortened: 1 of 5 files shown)


nurseiit:
Could anyone take a look please :pray: cc @mmstick and co

mmstick:
This page will be a port of the firmware manager application that's pre-installed in the COSMIC 24.04 ISO, which will display firmware update information from fwupd and system76-firmware.

nurseiit:
> This page will be a port of the firmware manager application that's pre-installed in the COSMIC 24.04 ISO, which will display firmware update information from fwupd and system76-firmware.

understood, thanks

PR #1235, opened 2025-06-16
i18n: fix window-management-appearance

Lampan-git:
Fix regression from #1230


Lampan-git:
@therealmate Do you want to check that this looks ok?

mmstick (approved):

PR #1237, opened 2025-06-18
feat: type to search

Doods0:
Type to search is a very convenient feature. It is often used in GNOME applications.

I implemented it by adding a new `Message` variant, if there was a way to send two messages at a time then I would've done that instead.


mmstick:
Created local test branch `tts`

Doods0:
>the search bar is labeled Type to search... but typing searches Settings now rather than the time zones.

Actually, type-to-search doesn't seem to work on any Context Drawer, even on master, so I'll just remove the labels.

I want to at least move the focus to the search bar when the drawer is opened, but I don't currently know how, so I'll delay it if you'll allow me.

Edit: type-to-search doesn't even work in the "Keyboard shortcuts" screen.
Edit 2: I didn't change the code releated to typing-to-search, I just messed with commits.
Edit 3: Commit 1eb3450 seems to have broken formatting, so the PR will temporarily include formatting changes .

jacobgkau (approved):
Keyboard navigation with Tab and Enter behave the same as they did before. Search sorting behaves the same as before. Typing into fields such as the hostname or WiFi password boxes still works. When typing a letter with no other input field focused, the search box is now focused automatically.

I did notice that when connecting to a WiFi network, the password dialog doesn't get focused automatically. Typing when it opens up without manually focusing it first now results in a search being done, whereas it used to result in nothing happening. This isn't a regression, but this change may make that bug more noticeable/annoying, so I've opened up https://github.com/pop-os/cosmic-settings/issues/1737 to track that one.

<details>
<summary>Regression testing passed:</summary>

## Checklist

### Network & Wireless

#### Wi-Fi

- [X] Able to disconnect from and connect to a network.

### Desktop

#### Wallpaper

- [X] Clicking a different wallpaper switches to it immediately
- [X] Leaving the page and coming back does not change the wallpaper

#### Appearance

- [X] Open a COSMIC app and a GNOME app. When switching between Dark and Light mode, the COSMIC app switches immediately, and the GNOME app switches after restarting.
- [X] Changing the accent color works.
- [X] Changing the roundness style works and does not affect other aspects of the theme (e.g. color mode).
- [X] Changing the interface density works.

#### Panel

- [X] Enable tiling. "Automatically hide panel" takes effect immediately.
    - Not a regression.

#### Dock

- [X] Enable tiling. "Automatically hide dock" takes effect immediately.
- [ ] Disable tiling; maximize a window. "Autommatically hide dock" takes effect immediately.
    - Not a regression.
- [X] Changing position on screen works.
- [X] All Style settings take effect as expected.

### Displays

- [X] Plug in a second display. Displays can be rearranged via dragging.

### Sound

- [X] Page works as expected.

### Power & Battery

- [X] Changing power mode takes effect and is reflected by top panel applet.
    - Change in applet doesn't reflect in Settings until re-entering the page; not a regression.

### Input Devices

#### Touchpad

- [X] All click behavior settings work as expected.
- [X] All scrolling settings work as expected.

### Applications

#### Default Applications

- [X] Changing the default terminal affects Super-T shortcut.

- [X] Global Shortcuts in X11 Applications works (in e.g. Discord).

### Time & Language

#### Date & Time

- [X] Changing time zone works.
- [X] Toggling 24-hour time takes effect in Settings preview, on top panel, and on the lock screen.
- [X] Toggling seconds takes effect in Settings preview and on top panel.
- [X] Changing first day of week takes effect in Calendar applet.
- [X] Turning date on/off for applet works.

### System & Accounts

#### About

- [X] Changing hostname works (check terminal prompt).
</details>

mmstick (approved):

PR #1245, opened 2025-06-24
Change wifi icon to one single icon

Lampan-git:
Change the icons to one instead of using two icons.
| Before | After |
| ------------- | ------------- |
| Screenshot_2025-06-25_18-03-51 | Screenshot_2025-06-24_18-48-33  |

### DIFF
--- a/cosmic-settings/src/pages/networking/wifi.rs
+++ b/cosmic-settings/src/pages/networking/wifi.rs
@@ -703,11 +703,10 @@ fn devices_view() -> Section<crate::pages::Message> {
                         };
 
                         let identifier = widget::row::with_capacity(3)
-                            .push(widget::icon::from_name(wifi_icon(network.strength)))
-                            .push_maybe(
-                                is_encrypted
-                                    .then(|| widget::icon::from_name("connection-secure-symbolic")),
-                            )
+                            .push(widget::icon::from_name(wifi_icon(
+                                network.strength,
+                                is_encrypted,
+                            )))
                             .push(
                                 widget::text::body(network.ssid.as_ref()).wrapping(Wrapping::Glyph),
                             )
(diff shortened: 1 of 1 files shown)


git-f0x:
This doesn't technically match [designs](https://www.figma.com/design/Z6Z9Ox6WNkiiM8gcAkcElS/COSMIC-Settings-First-Release---Developer-Handoff?node-id=6922-151228&t=Yb6QmHWwnTrNpVub-0), but IMO, it looks quite a bit better.
Though there are maybe potential accessibility concerns, so needs UX approval regardless.

Lampan-git:
That's fair. I just want to note that these icons seems to have been added only 10 months ago, which is probably after the design document.

mmstick (approved):

maria-komarova:
Where are the new icons coming from?
There were a couple small concerns with them using the lock like that:
- I worry the lock is too small to be easily recognizable, at least for some people. Although familiarity of that kind of placement should help a little. But the concern remains, the lock image gets too small. 
- Wi-fi icon without a lock looks misaligned compared to other icons if used next to them. This happens because of the free space allocated for the lock image. So we'd need to make sure the icon is not used anywhere in that context or do something about the icon. 

Those concerns are the main reason the icons have not been changed to what is suggested here earlier.

mmstick:
We should drop the icon patch for now, and move it to its own PR then.

Lampan-git:
> Where are the new icons coming from?

https://github.com/pop-os/cosmic-icons/blob/master/extra/scalable/status/network-wireless-secure-signal-ok-symbolic.svg

> We should drop the icon patch for now, and move it to its own PR then.

I created a new one for the open wifi instead #1248, since the icon discussion is here.

git-f0x:
>Wi-fi icon without a lock looks misaligned compared to other icons if used next to them. This happens because of the free space allocated for the lock image. So we'd need to make sure the icon is not used anywhere in that context or do something about the icon.

This also happens with the network applet, since the WiFi icon ends up looking misaligned compared to other applet icons. So the icons should possibly be tweaked to be centered with and without the lock icon.

PR #1258, opened 2025-06-29
Add Git SHA support and  argument parsing for --version and --help

LinuxBoy-96:
Cosmic Settings is next in line to support argument parsing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1671,6 +1671,7 @@ dependencies = [
  "bluez-zbus",
  "chrono",
  "clap",
+ "clap_lex",
  "color-eyre",
  "cosmic-bg-config",
  "cosmic-comp-config",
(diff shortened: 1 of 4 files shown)


mmstick:
The cosmic-settings application will always be launched from a desktop entry, so there's no use for these arguments. I'm trying to reduce dependencies and code generation, so you could help out by refactoring the current argument parsing to use clap_lex instead of clap.

LinuxBoy-96:
> The cosmic-settings application will always be launched from a desktop entry, so there's no use for these arguments. I'm trying to reduce dependencies and code generation, so you could help out by refactoring the current argument parsing to use clap_lex instead of clap.

So just because cosmic-settings is not supposed to be lauched from the terminal (btw I do it just because I can), It don't deserve a proper --version and --help support?

I would gladly add migrating other parts that currently use clap to clap_lex to my TODO list, should this change be accepted. I see it as a natural continuation of this integration.

My code did indeed introduce Clap_lex.

mmstick:
Yes, this application only uses argument parsing for implementing desktop entries for each page in the application.

PR #1260, opened 2025-06-30
From Clap to Clap_lex

LinuxBoy-96:
@mmstick mentioned to me that he would like to see clap_lex instead of clap, so... here it is.

To my knowledge, it works perfectly.

The only bad news is that I removed the cfg attributes from individual functions, as they made the code really hard to read.
I think there are better ways to handle specific functionality when it's not applicable to a platform than doing it at compile time.

Feel free to hate it, but please don't leave me in the dark.

### DIFF
--- a/cosmic-settings/src/main.rs
+++ b/cosmic-settings/src/main.rs
@@ -21,122 +21,119 @@ pub mod theme;
 pub mod utils;
 pub mod widget;
 
-use clap::{Parser, Subcommand};
+use clap_lex::RawArgs;
 use cosmic::{app::CosmicFlags, iced::Limits};
 use i18n_embed::DesktopLanguageRequester;
-use ron::error::SpannedError;
 use serde::{Deserialize, Serialize};
 use tracing_subscriber::prelude::*;
 
-#[derive(Parser, Debug, Serialize, Deserialize, Clone)]
-#[command(author, version, about, long_about = None)]
-#[command(propagate_version = true)]
+
 pub struct Args {
-    #[command(subcommand)]
     sub_command: Option<PageCommands>,
 }
 
-#[derive(Subcommand, Debug, Serialize, Deserialize, Clone)]
+impl Args {
+    pub fn parse() -> Self {
+        let raw_args = RawArgs::from_args();
+        let mut cursor = raw_args.cursor();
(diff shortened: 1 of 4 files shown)


cosmic-settings/src/main.rs
```diff
@@ -21,122 +21,119 @@ pub mod theme;
...
-    /// Wired settings page
-    #[cfg(feature = "page-networking")]
     Wired,
-    /// WiFi settings page
-    #[cfg(feature = "page-networking")]
     Wireless,
-    /// Workspaces settings page
-    #[cfg(feature = "page-workspaces")]
     Workspaces,
 }
```
mmstick, line 85: You must not remove the conditional compilation annotations

mmstick (changes requested):

cosmic-settings/src/main.rs
```diff
@@ -21,122 +21,119 @@ pub mod theme;
...
+            "System" => Some(Self::System),
+            "Time" => Some(Self::Time),
+            "Touchpad" => Some(Self::Touchpad),
+            "Users" => Some(Self::Users),
+            "Vpn" => Some(Self::Vpn),
+            "Wallpaper" => Some(Self::Wallpaper),
+            "WindowManagement" => Some(Self::WindowManagement),
+            "Wired" => Some(Self::Wired),
+            "Wireless" => Some(Self::Wireless),
+            "Workspaces" => Some(Self::Workspaces),
```
mmstick, line 120: This is not in the same format as clap was using before. Make sure every desktop entry works.

LinuxBoy-96: @mmstick By that you mean only conditionnal compilation annotation?

LinuxBoy-96: > @mmstick By that you mean only conditionnal compilation annotation? If not I tested it... this if the Clap_lex way to do it. :p

mmstick: No, our desktop entries are launched via the name of the page, but it is not written in the same way as in the source code. `WindowManagement` is `Exec=cosmic-settings window-management`, for example. So again, make sure the desktop entries work with the change.

PR #1267, opened 2025-07-07
i18n (ar): Updated Arabic translation.

defaultUser822:


Doods0:
I think that's better, if my opinion even matters.

jermanuts:
Hello, I updated the whole translation file including this string, feel free to review it

https://github.com/pop-os/cosmic-settings/pull/1345

Issue #1264, opened 2025-07-04
[Bug] Wallpaper flickers to a different one if COSMIC Settings was launched straight into the "Desktop>Wallpaper" page


abdu27th:
### Summary
The wallpaper changes to a different one for a split second and back to the set one. This happens when COSMIC Settings is launched straight into the Desktop>Wallpaper page. The flickering happens if the wallpaper is chosen from the default "Wallpapers" or "Colors" folders. It seems to be flickering to the default wallpaper (orion_nebula_nasa_heic0601a.jpg) every time.

---

### Steps to Reproduce
1. Launch COSMIC Settings and navigate to Desktop > Wallpaper page.
2. Set the wallpaper to one in the "Wallpapers" or "Colors" folder.
3. Close and re-launch COSMIC Settings.
4. Observe the flickering of wallpapers to a different one and back to the set one.

### Environment
- OS: Pop!_OS 24.04 LTS

Doods0:
As-salamu alaykum.

I confirm this happens on master, but it seems to be flicking into a blank color instead of another wallpaper.

PR #1274, opened 2025-07-13
fix(wallpaper): Wallpaper flickering when starting into the wallpaper page.

Doods0:
Fixes: #1264

I just copied the "fix" code from `WallpaperEvent::Load` to `WallpaperEvent::Loaded`, so it only runs after we finish loading wallpapers, or at least this is what I think I did. The change isn't obvious when viewing the diff.


Doods0:
No, I think I made a mistake.

Issue #1293, opened 2025-07-26
[Fedora] Polkit rules reference the wrong sudoers group


zarugo:
Since this is distro-specific. it's most likely a packaging concern, but since the polkit rule files are in this repo, those rules are not correct in a fedora-based distro, where the default group for sudoers is "wheel" and not "sudo".

`subject.isInGroup ("sudo")`

 should normally be 

`subject.isInGroup ("wheel")`

I am not sure if this, at the moment, has any real consequences. I did some tests trying to perform the actions where this rule applies (like changing the keyboard layout). I expected an authentication request instead of the action just working, but it just worked . Not sure if this is taken care of by some other polkit rule in the system, I did not investigate further.

The only rules I see belonging to a cosmic package are

```
cosmic-settings-daemon.rules
cosmic-settings.rules
```

mmstick:
You'll want to apply the patch to the Fedora packaging

PR #1294, opened 2025-07-26
fix: also accept "wheel" as group name for polkit authentication

therealmate:
Fixes https://github.com/pop-os/cosmic-settings/issues/1293

### DIFF
--- a/resources/polkit-1/rules.d/cosmic-settings.rules
+++ b/resources/polkit-1/rules.d/cosmic-settings.rules
@@ -6,7 +6,7 @@ polkit.addRule(function(action, subject) {
             action.id == "org.freedesktop.hostname1.set-hostname") &&
         subject.local &&
         subject.active &&
-        subject.isInGroup ("sudo")) {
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
Before this change, after removing a Pop!_OS 24.04 user from the `sudo` group and adding it to a new `wheel` group, changing the hostname ("Device name" setting on the About page) didn't work. After this change, I'm able to change the hostname while being part of `wheel` and not `sudo`.

Note that I didn't actually change the sudoers group from `sudo` to `wheel` during this test (and was thus unable to run `sudo` commands on the terminal), so this isn't actually checking if the user has root permissions, it's only checking if they're part of one of these group names. But since both of these names are de-facto reserved for sudoers use, and new groups can't be added without root privileges in the first place (at least on Pop!_OS), this doesn't seem like a problem. (If it was a security concern for Pop!_OS, then non-Pop!_OS distros using group names other than `sudo` would already have the same problem right now.)

mmstick (approved):

PR #1306, opened 2025-08-02
fix searching for input source switch keyboard shortcut in keyboard shortcut page

voedipus:


voedipus:
Before:

Screenshot_2025-08-08_14-21-04


After:

Screenshot_2025-08-08_14-16-54

Issue #1346, opened 2025-09-07
[BUG] Misiing 'Hardware model' and 'Processor' information in the About section


cappsyco:
While testing on a M1 Macbook running Asahi Fedora Remix, I noticed the model and CPU information does not show in settings. This is due to the relevant information not being present in /proc/cpuinfo or /sys/devices/virtual/dmi/id/

<img width="1766" height="1181" alt="Image" src="https://github.com/user-attachments/assets/b1f2d021-5e70-49b0-9312-ea880af8fd54" />

PR incoming to provide a fallback using the already existing sysinfo crate but due to the nature of this it's very difficult to confirm if this will cover all eventualities.

PR #1347, opened 2025-09-07
fix: adding fallback for Hardware Model and Processor in the About section

cappsyco:
Addresses issue https://github.com/pop-os/cosmic-settings/issues/1346

This provides fallback values for Hardware Model and Processor, using the sysinfo crate that's already used for other values on the page. This should guarantee no blank values, but currently I've only been able to test on a M1 Macbook Air running Asahi Fedora.

Screenshot_2025-09-08_00-22-52

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1709,6 +1709,7 @@ dependencies = [
  "smithay-client-toolkit",
  "static_init",
  "sunrise",
+ "sysinfo 0.37.0",
  "tachyonix",
  "timedate-zbus",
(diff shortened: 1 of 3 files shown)


mmstick:
Make sure to run `cargo fmt`. May want to set your IDE to auto format on save.

cappsyco:
> Make sure to run `cargo fmt`. May want to set your IDE to auto format on save.

Zed was obviously having a moment! Sorted now

PR #1376, opened 2025-09-20
i18n(bg): updated bulgarian translation

twlvnn:


twlvnn:
Thats so cool, thank you for making a weblate instance!!!

PR #1380, opened 2025-09-22
Update cosmic_settings.ftl ITA

lucamosca1:


mmstick:
We're on weblate now: https://hosted.weblate.org/projects/pop-os/

PR #1423, opened 2025-09-29
i18n(be): Update belarusian translation

aliaksandr-trush:


mmstick:
Translations need to be done through weblate now: https://hosted.weblate.org/projects/pop-os/

PR #1460, opened 2025-10-04
fix: case sensitivity in applet search

cappsyco:
Addresses https://github.com/pop-os/cosmic-settings/issues/1456


cappsyco:
For some reason my rustfmt changed things in a way the CI didn't like. Fixed now.

mmstick:
Use this instead. https://crates.io/crates/unicase

You'll be able to search without allocating new strings with this.

cappsyco:
Sorry, I'm coming up short when trying to get unicase working with substring matching. It works great for exact match comparison but I think that's not really suitable for this search function.

I've switched out to_lowercase with the ascii version for performance improvements while still doing the allocations for now while I dig a bit more

PR #1470, opened 2025-10-08
Remove -d '\n' from xargs

libkurisu:
This fixes unrecognized option d errors on
freebsd utils.

This is unnecessary as \n is already a default limiter for Xargs and it breaks building with the BSD utils.
I tried to get this working with find -print0 and xargs -0 without any luck but given that there are no files with a space or newline in it's name (and that the current script would break anyways if there was one) I think this should be fine :)


mmstick (approved):

mmstick:
Thanks. I'll be streamlining the justfile soon.

PR #1475, opened 2025-10-09
chore: update libcosmic

wash2:
fix for color picker


mmstick:
Seems cosmic-dbus-networkmanager needs to be updated to get the lockfile to procfs 0.18

Issue #1325, opened 2025-08-22
Sound page update


WatchMkr:
Lab testing found that attempting to attach profiles to "Output" and "Input" was often confusing, especially when multiple devices are available, and sometimes resulted in broken audio setups.

* Separate profiles to a "Device Profiles" page
* Include the port in the Out/Input Device name "Port" - "Device" (e.g. "Headphones - Built-in Audio")

<img width="870" height="618" alt="Image" src="https://github.com/user-attachments/assets/9bdfd608-fa4c-4184-b21f-d53a75021c69" />

<img width="872" height="485" alt="Image" src="https://github.com/user-attachments/assets/6ff72996-f690-49c2-bf5d-4824fc050cc1" />

KodeBarista:
Constructive criticism: Slider design looks terrible (not in a good way)

<img width="704" height="169" alt="Image" src="https://github.com/user-attachments/assets/3eb13e0e-91d1-4b46-a82d-aab60257774b" />

Current design already looks better, maybe it needs more space between the text & slider but everything else looks great. 
Why can't just move the number 100 to the top of the vertical line in the current design?

In the current design we can see that there's inconsistency in the spacing between slider & text. 
- Slider 1 with value "40" has sufficient space.
- Slider 2 with values (left & right) doesn't have the same amount of space.
- Slider 3 also lacks the same amount of space as slider 1. 

<img width="770" height="584" alt="Image" src="https://github.com/user-attachments/assets/6dc89174-4ee5-4135-afb0-9fe913313a17" />

PR #1500, opened 2025-10-20
feat(sound): redesign with separate device profiles page

mmstick:
Closes pop-os/cosmic-settings#1325 
Closes pop-os/cosmic-osd#158 
Closes https://github.com/pop-os/cosmic-epoch/issues/655
Closes https://github.com/pop-os/cosmic-epoch/issues/787
Closes https://github.com/pop-os/cosmic-epoch/issues/946
Closes https://github.com/pop-os/cosmic-epoch/issues/1001
Closes https://github.com/pop-os/cosmic-epoch/issues/568
Closes https://github.com/pop-os/cosmic-epoch/issues/1368

- Device profiles have been moved into a new Device Profiles sub-page for selecting profiles per device
    - Set with  `wpctl set-profile {{device-id}} {profile-index}}` instead of `pactl`
- Setting the default sink (output) and source (input) now operates independently
    - Set with `wpctl set-default {{node-id}}`
- All of our devices now have stable node object and device IDs from pipewire-rs
        - Enables using wpctl for setting card profiles, default sinks/sources, volume, and mute
        - Significant performance and stability improvements
        - Wireplumber state no longer gets corrupted
- Fixed early return on debounced volume events dropping pipewire and pulse events
- Update pipewire-rs to version 0.9
- Most of our dependency on libpulse has been eliminated
    - It is only used for getting and setting volume, volume balance, and watching default sink/source changes

### DIFF
--- a/cosmic-settings/src/pages/sound/mod.rs
+++ b/cosmic-settings/src/pages/sound/mod.rs
@@ -78,7 +78,6 @@ pub struct Page {
 
 impl Default for Page {
     fn default() -> Self {
-        fix_wireplumber_stream_properties();
         let mut model = subscription::Model::default();
         model.unplugged_text = fl!("sound-device-port-unplugged");
         model.hd_audio_text = fl!("sound-hd-audio");
@@ -458,30 +457,6 @@ fn device_profiles() -> Section<crate::pages::Message> {
 
             settings::section().add(device_profiles).into()
         })
-}
-
-fn fix_wireplumber_stream_properties() {
-    let path = std::env::home_dir()
-        .expect("no home dir")
-        .join(".local/state/wireplumber/stream-properties");
-
-    let Ok(mut data) = std::fs::read_to_string(&path) else {
-        return;
-    };
-
-    let mut insert_pos = 0;
-    let mut lines = data.split('\n');
-    while let Some(line) = lines.next() {
-        if line.starts_with("Input/Audio:") {
-            return;
-        } else if line.starts_with("Output") {
-            break;
-        }
-
-        insert_pos += 1 + line.len();
-    }
-
-    data.insert_str(insert_pos, "\nInput/Audio:application.id:org.PulseAudio.pavucontrol={\"channelMap\":[\"MONO\"], \"mute\":false, \"volume\":1.000000, \"channelVolumes\":[1.000000]}\"");
 }
 
 // fn alerts() -> Section<crate::pages::Message> {


mmstick:
What's the output of `pw-dump` on this system? This is what I see on a system with a Realtek card, Radeon HDMI audio, and a USB-C sound card.

Screenshot_2025-10-20_19-25-58

Screenshot_2025-10-20_19-27-00

mmstick:
Fixed the devices on the device profiles page disappearing when setting their profile to off. Which also fixes the device briefly disappearing when the device's nodes are replaced.

jacobgkau (changes requested):
On the same machine as yesterday, I still had one of the devices not have any profile visibly selected at first, which shouldn't be possible:

Screenshot_2025-10-21_11-21-24

Here is `pw-dump` from that system: [pw-dump-oryp8.txt](https://github.com/user-attachments/files/23027547/pw-dump-oryp8.txt)

If I set the device to "Off", it initially shows itself as "Off", but if I reboot, I end up in the same state as shown above (it boots up set to nothing instead of showing as set to "Off".)

jacobgkau:
I'm still also seeing menu items getting cut off at the edge of the window. Is that a libcosmic bug? Dropdown menus in other COSMIC apps seem to be able to run out the side of the window.

mmstick:
It's a different dropdown widget. I will switch to it

mmstick:
The index of the selected profile on that Built-in Audio device should be fixed now

jacobgkau:
On 8688b54 and oryp8, the profile of `Built-in Audio` keeps defaulting to `Digital Surround 5.1 (IEC958/AC3) Output`. I think this corresponds to the S/PDIF output, but attempting to change the volume does nothing. Going directly to `Digital Stereo (IEC958) Output` doesn't work, but if I choose `Off`, then I can go to `Digital Stereo (IEC958) Output` to get sound out of the S/PDIF port (the `Digital Surround` option then gives distorted audio if I select that again).

If I just boot (defaulting to `Digital Surround 5.1 (IEC958/AC3) Output`) and change directly to one of the `Analog Stereo` options, that also doesn't work (changing the volume doesn't work & there's no sound). Again, I have to select `Off` first, then go back to `Analog Stereo` for it to work.

I know some of this might be audio card weirdness that's outside the scope of COSMIC Settings, but does any of this sound like a COSMIC Settings thing? The two distinct problems are the default/last selected profile not being saved between sessions, and not being able to switch from the default to one of the others successfully without first selecting `Off` (or possibly just sound not working at all until it's turned `Off`).

jacobgkau:
From @WatchMkr in Mattermost (posting here so it doesn't get lost)-- the design also calls for the device description to be prepended to the profile name when selecting an input/output device (e.g. `Speakers - Built-in Audio`). The string to be prepended appears to be labeled "description" in `pw-dump`.

mmstick:
Seems I'll need to add the different routes (ports) to the output/input dropdown lists. Pavucontrol puts those on a separate port dropdown.

jacobgkau:
Not sure if you were ready for me to look at this again yet, but just to corroborate, on f787c83, I'm seeing the `Output device` as `Analog Stereo - Built-in Audio`, when it should say `Speakers - Built-in Audio`. As you noted, it should be the description of the route, not the profile.

The other problems I mentioned before are still occurring.

mmstick:
What's the output of `pw-dump` when set to `Analog Stereo Duplex`? Particularly the `Pipewire:Interface:Node` associated with it. Changing the profile changes the active route, and a  `PipeWire:Interface:Node` is created from that.

Based on the logic that I have, the most likely problem is that the node failed to parse. The node ID is required to set a default sink/source, and the route won't be added to the device name if a node isn't associated with it.

mmstick:
I think I see the issue

mmstick:
Hoping this fixes the issue you're seeing.

jacobgkau (commented):
On 0e7d8b2, when I switch the profile to `Analog Stereo Duplex`, the output device and input device are labeled correctly (`Speakers - Built-in Audio` and `Internal Microphone - Built-in Audio`, respectively). However, the volume hotkeys don't work right away.

If I open up Firefox and start playing sound, the OSD pops up by itself, and then the hotkeys start working. Alternatively, if I switch to `Analog Stereo Duplex` (with no working hotkeys), then switch to `Analog Stereo Output`, the hotkeys start working, and continue working when I switch back to `Analog Stereo Duplex`.

- pw-dump after switching to `Analog Stereo Duplex`, with hotkeys not working: [oryp8-0e7d8b2-duplex-nohotkeys.txt](https://github.com/user-attachments/files/23243531/oryp8-0e7d8b2-duplex-nohotkeys.txt)
- pw-dump after playing audio in Firefox, when the hotkeys are working: [oryp8-0e7d8b2-duplex-soundandhotkeys.txt](https://github.com/user-attachments/files/23243534/oryp8-0e7d8b2-duplex-soundandhotkeys.txt)

The profile is also still defaulting to `Digital Surround 5.1 (IEC958/AC3) Output` every time I boot up.

mmstick:
The only difference I see in the pw-dump is the ALC1220 playback_FL port increasing its AsyncBuffers from size 8 to 16, and the addition of a Firefox client/link objects using that port. We can try setting the default with pactl and wpctl simultaneously. Seems like a quirk with pipewire itself not advertising the profile or sink change.

jacobgkau:
On f5209ce, still getting the quirk of volume hotkeys not working immediately after switching the profile to `Analog Stereo Duplex`. Also still defaulting to `Digital Surround 5.1 (IEC958/AC3) Output` on every boot.

Just to confirm, I went back to master and was still seeing it default to the digital output (with no other profiles even showing up as available). I installed `pavucontrol` and switched the profile with that, and then it defaulted to the correct profile on the next boot. After installing this branch again, it also defaults to the correct profile.

It seems like maybe some bad PulseAudio/PipeWire config may've been stored that `pavucontrol` was able to wipe out? I think I'm going to do a clean install on the system to see if I can replicate the problems I was seeing before. (It would still be good for COSMIC Settings to be able to handle whatever was going on like `pavucontrol` was able to.)

jacobgkau:
After doing a fresh install from the 24.04 20 ISO, then adding this branch, the volume hotkeys work fine. The system does default to `Analog Stereo Duplex` on every boot (e.g. if I change it to `Analog Stereo Output` and reboot, that's not saved), but `Analog Stereo Duplex` is at least the best one for the oryp8).

After going back to the currently released version of COSMIC Settings (a8d8e14), I rebooted and then messed with the profiles in the Input and Output section a bunch, then installed this branch and rebooted again. That resulted in the broken state I described earlier. So this branch of COSMIC Settings is not able to correct something the previous design was doing wrong, while `pavucontrol` is able to correct it.

I'll try to narrow down specific steps to cause the breakage with the old design before upgrading to this version.

jacobgkau:
I can fix the issue by removing `~/.local/state/wireplumber/` and `~/.config/pulse/`, then rebooting.

I seem to be able to trigger the broken state by taking the following steps:

1. Revert to the master branch
2. Set the `Output` profile to `Digital Surround 5.1 (IEC958/AC3) Output`
    - When I do this, the Input section disappears, and the Output device's name changes to `Built-in Audio Digital Surround 5.1 (IEC958/AC3)`, with the Output profile setting being blank.
3. Install this branch and reboot.

The profile then defaults to `Digital Surround 5.1 (IEC958/AC3) Output` (which is somewhat expected), and the volume hotkeys don't work immediately if I switch to `Analog Stereo Duplex` (which is a bug). Opening `pavucontrol` immediately plays a volume adjustment blip, and then the hotkeys work again.

---

This is the contents of `~/.local/state/wireplumber/stream-properties` when the profile is set to `Digital Surround 5.1 (IEC958/AC3) Output` with the old design:

```
[stream-properties]
Audio/Sink:node.name:alsa_output.pci-0000_00_1f.3.iec958-ac3-surround-51={"channelVolumes":[1.000000, 1.000000, 1.000000, 1.000000, 1.000000, 1.000000], "channelMap":["FL", "FR", "RL", "RR", "FC", "LFE"]}
Audio/Sink:node.name:auto_null={"volume":1.000000, "channelMap":["FL", "FR"], "channelVolumes":[1.000000, 1.000000], "mute":false}
Output/Audio:media.role:Music={"volume":1.000000, "channelMap":["FL", "FR"], "channelVolumes":[1.000000, 1.000000], "mute":false}
```

This is the contents of that file in the broken state, after changing the output device to Analog Stereo Duplex with this branch:

```
[stream-properties]
Audio/Sink:node.name:alsa_output.pci-0000_00_1f.3.iec958-ac3-surround-51={"channelMap":["FL", "FR", "RL", "RR", "FC", "LFE"], "channelVolumes":[1.000000, 1.000000, 1.000000, 1.000000, 1.000000, 1.000000]}
Audio/Sink:node.name:auto_null={"volume":1.000000, "mute":false, "channelMap":["FL", "FR"], "channelVolumes":[1.000000, 1.000000]}
Output/Audio:media.role:Music={"volume":1.000000, "channelMap":["FL", "FR"], "channelVolumes":[1.000000, 1.000000], "mute":false}
```

This is the contents immediately after opening `pavucontrol` (not even touching any settings), which is when the hotkeys start working again:

```
[stream-properties]
Audio/Sink:node.name:alsa_output.pci-0000_00_1f.3.iec958-ac3-surround-51={"channelMap":["FL", "FR", "RL", "RR", "FC", "LFE"], "channelVolumes":[1.000000, 1.000000, 1.000000, 1.000000, 1.000000, 1.000000]}
Audio/Sink:node.name:auto_null={"volume":1.000000, "mute":false, "channelMap":["FL", "FR"], "channelVolumes":[1.000000, 1.000000]}
Input/Audio:application.id:org.PulseAudio.pavucontrol={"channelMap":["MONO"], "mute":false, "volume":1.000000, "channelVolumes":[1.000000]}
Output/Audio:media.role:Music={"volume":1.000000, "mute":false, "channelMap":["FL", "FR"], "channelVolumes":[1.000000, 1.000000]}
```

All other config files in `~/.local/state/wireplumber/` and `~/.config/pulse/` are identical between the broken state and the fixed state.

@mmstick Does this help see what else COSMIC Settings needs to do when setting the profile in order to handle this kind of situation?

mmstick:
No, added the missing `--user`.

jacobgkau:
On c8880c8, I see `wireplumber.service` being restarted the first time I open Sound Settings in the broken state, so that part's working. However, the volume hotkeys don't start working after that. If I attempt to change the profile from the `Digital Surround 5.1 (IEC958/AC3) Output` profile it defaults to in the broken state to `Analog Stereo Duplex`, the volume hotkeys still don't start working. They still start working right away if I open up `pavucontrol`.

This is what `stream-properties` looks like when booting up in the broken state:

```
[stream-properties]
Audio/Sink:node.name:alsa_output.pci-0000_00_1f.3.iec958-ac3-surround-51={"channelMap":["FL", "FR", "RL", "RR", "FC", "LFE"], "channelVolumes":[1.000000, 1.000000, 1.000000, 1.000000, 1.000000, 1.000000]}
Audio/Sink:node.name:auto_null={"volume":1.000000, "mute":false, "channelMap":["FL", "FR"], "channelVolumes":[1.000000, 1.000000]}
Output/Audio:media.role:Music={"channelVolumes":[1.000000, 1.000000], "volume":1.000000, "mute":false, "channelMap":["FL", "FR"]}
```

This is what it looks like after opening Sound Settings for the first time, with the hotkeys still not working:

```
[stream-properties]
Audio/Sink:node.name:alsa_output.pci-0000_00_1f.3.iec958-ac3-surround-51={"channelVolumes":[1.000000, 1.000000, 1.000000, 1.000000, 1.000000, 1.000000], "channelMap":["FL", "FR", "RL", "RR", "FC", "LFE"]}
Audio/Sink:node.name:auto_null={"mute":false, "volume":1.000000, "channelMap":["FL", "FR"], "channelVolumes":[1.000000, 1.000000]}
Input/Audio:application.id:org.PulseAudio.pavucontrol={"channelMap":["MONO"], "mute":false, "volume":1.000000, "channelVolumes":[1.000000]}
Output/Audio:media.role:Music={"channelVolumes":[1.000000, 1.000000], "volume":1.000000, "mute":false, "channelMap":["FL", "FR"]}
```

This is what it looks like after opening `pavucontrol`, when the hotkeys start working:

```
[stream-properties]
Audio/Sink:node.name:alsa_output.pci-0000_00_1f.3.iec958-ac3-surround-51={"channelVolumes":[0.857375, 0.857375, 0.857375, 0.857375, 0.857375, 0.857375], "channelMap":["FL", "FR", "RL", "RR", "FC", "LFE"]}
Audio/Sink:node.name:auto_null={"mute":false, "volume":1.000000, "channelMap":["FL", "FR"], "channelVolumes":[1.000000, 1.000000]}
Input/Audio:application.id:org.PulseAudio.pavucontrol={"mute":false, "channelMap":["MONO"], "volume":1.000000, "channelVolumes":[1.000000]}
Output/Audio:media.role:Music={"channelVolumes":[1.000000, 1.000000], "volume":1.000000, "mute":false, "channelMap":["FL", "FR"]}
```

They look pretty similar, so I'm not sure the problem's in there.

After replicating the issue again, I also tried opening up this branch and then manually restarting `pipewire.service` and `pipewire.socket`, which didn't help.

mmstick:
I looked around the Internet for random bug reports about this. There are some mixed results. Some mention that muting and unmuting may fix the device. Others report than changing the profile fixes it. One person mentioned needing to restart `pipewire-pulse`.

Pavucontrol's source code was quite cryptic to read through, but it doesn't seem to be doing anything special besides re-setting the current settings? So I added some wpctl calls to run `wpctl set-profile {{device_id}} {{profile_index}}` and `wpctl set-default {{node_id}}` when we get notice of a new profile and device default. Maybe there's a better way, but it feels like a bug in wireplumber or pipewire.

mmstick:
If you run `pw-cli ls Node` you can get a list of nodes and their IDs. `pw-cli ls Device` for devices. `pw-cli e {{device-id}} Profile` to get the active profile and its index for a device (albeit the output is not very human-friendly). It may also be worth trying out `pw-cli s {{device-id}} Profile '{ index: {{profile-index}}, save: true }'` from https://gitlab.freedesktop.org/pipewire/pipewire/-/wikis/Migrate-PulseAudio#set-card-profile

jacobgkau:
On 1b93fdf, when I opened the new settings page for the first time after messing up the profile with the old one, the repair defaulted it to having the device's profile set to `Off`. Changing it to `Analog Stereo Duplex` got me working hotkeys immediately. I think this is acceptable behavior.

However, when I rebooted the system, it defaulted to `Digital Surround 5.1 (IEC958/AC3) Output`. Then, changing it to `Analog Stereo Duplex` has broken hotkeys again. What's responsible for saving/loading the config between sessions?

I'm attempting to compare the config using the `pw-cli` commands you suggested.

jacobgkau:
Using `pw-cli`, I discovered that the `Analog Stereo Duplex` profile has an index of `1`. (`off` is index `0`, and the bad 5.1 digital profile is index `30`.)

Coming from the bad state, if I run `pw-cli s 49 Profile '{ index: 1, save: true }'`, then reboot, the profile's still set to `1` as I'd expect. But if I instead open COSMIC Settings, first of all, the device ID changes from 49 to 51, and then, if I change to `Analog Stereo Duplex` in COSMIC Settings, the profile's set to `1` for the time being, but goes back to `30` after rebooting.

The `Prop: key Spa:Pod:Object:Param:Profile:save` value is set to `false` when COSMIC Setting sets it, but obviously `true` when `pw-cli` sets it with `save: true` included. That's probably part of the issue with saving across reboots.

mmstick:
The profile is now being set with `pw-cli`.

jacobgkau:
Still testing, but seems to be working better now. After recreating the broken state, the new version now switches profiles correctly (with hotkeys working immediately after the switch), and then loads the profile I switched to again on the next boot.

mmstick:
Yeah I'll add that

jacobgkau (commented):
83eecd5 didn't seem to work for that-- the text is still not updating until I close & re-open after plugging or unplugging headphones.

Another thing I noticed is that when I use the `Pro Audio` profile, all of the items in the output drop-down are the same (e.g. I have five drop-down items all labeled `Built-in Audio`). pavucontrol at least has numbers for each of them so they can be differentiated. It looks like pavucontrol displays the `node.description`, which shows up as e.g. `Built-In Audio Pro`, `Built-in Audio Pro 3`, `Built-in Audio Pro 7`, etc, instead of the ALSA card name.

mmstick:
Switched to using libpulse for monitoring active port changes, and I'm seeing it switch between Speakers and Headphones on my system. Checking the Pro Audio issue now.

jacobgkau:
On ff15c0d, the active port change is working. `Pro Audio` isn't necessarily showing the expected labels-- this is what I see in COSMIC Settings vs. pavucontrol:

Screenshot_2025-11-10_10-48-49

The speaker output actually being labeled Speakers is actually better than pavucontrol, but it still seems like it makes less sense for the rest to not have any identifiers. Again, it looks like we could get e.g. `Built-in Audio Pro 1` from `node.description`, or `Pro 1` from `device.profile.description`. If we don't want to just always use the `node.description` instead of `device.description`, then maybe we could use the `device.profile.description` in place of the route `description` when the latter's empty? That would make it show up as e.g. `Pro 1 - Built-in Audio`.

---

The `Confirm device type` pop-up is sometimes slow to trigger, and that causes issues in the Sound settings panel if I unplug before it's processed. E.g.:

- Plug in a headset; sound settings immediately changes `Output device` to reflect the output change.
- Unplug the headset.
- A few seconds later, the `Confirm Device Type` pop-up finally shows up.
- I choose `Headset` (even though I've already unplugged the headset) and click `Confirm`.
- Sound settings changes `Input device` to `Speakers - Built-in Audio`, which doesn't make sense.

Would this be a cosmic-osd issue?

mmstick:
Do you have a pw-dump with those devices on the pro audio profile?

mmstick:
Yeah the confirm device pop-up may be a cosmic-osd issue.

jacobgkau:
Here's a pw-dump while set to Pro Audio on oryp8: [oryp8-pro-audio.txt](https://github.com/user-attachments/files/23459901/oryp8-pro-audio.txt)

jacobgkau (commented):
Pro Audio labels are showing up as expected now.

Last thing I'm noticing is an issue with the Balance setting. It works if I'm playing audio and I change it. However, it resets to the middle every time I adjust the volume (via hotkey or slider). It should probably be unaffected by volume changes (although some way to reset it might be nice, since its slider doesn't snap to the middle).

mmstick:
Finally resolved. I'll need ~15-30 minutes to get this ready to commit though. We will be able to get and set channel volumes and other parameters directly in cosmic-settings without needing libpulse, wpctl, or pw-cli now. Now that we're subscribing to pipewire for parameter updates, I can drop libpulse as a dependency.

Updating the audio applet to share the same cosmic-settings library will ensure that our balance stays intact when using that as well. Changing the volume by shortcut uses wpctl though, which will overwrite the balance. I think we'll need to use cosmic-settings-daemon after this is merged to add an API for the shortcuts to use that preserves our balance.

jacobgkau (changes requested):
The latest update (to 9c51fff) broke the [test case](https://github.com/pop-os/cosmic-settings/pull/1500#issuecomment-3474548675) upgrading from the current stable version. I'm back to having no volume hotkeys after changing the profile to `Analog Stereo Duplex` on the new version.

Also, dragging the slider to adjust the volume doesn't seem to work anymore. It's reflected in the COSMIC Settings interface, but not pavucontrol or the OSD, and adjusting the volume with the hotkeys resumes from wherever it was before, not from where I set it to in COSMIC Settings.

mmstick:
Re-added the use of `pw-cli` in case that does something to fix this that a direct call to pipewire doesn't.

mmstick:
Not able to replicate any issues with volume sliders, so I suppose fixing the broken state will also fix that too.

mmstick:
Should be fixed now.

jacobgkau:
I noticed that if I e.g. plug in an HDMI cable while the `Device profiles` page is open, the list of profiles doesn't update to let me select it until I leave the page and re-enter it. Conversely, if I unplug HDMI while that page is open, the profiles that are no longer available don't disappear from the list until I leave the page & re-enter it. Is that something that could be refreshed when it changes or otherwise watched? (pavucontrol seems to update that info immediately in its list.)

mmstick:
Checking it. That is likely a simple fix

jacobgkau:
Hotplugging/unplugging entire devices now updates their available profiles mostly as expected. There are still a couple of quirks:

- If a drop-down is open, the profiles aren't updated until it's closed.
    - Would automatically closing the drop-down (if not re-opening it automatically) when there's an update be feasible? Not sure if this would require deeper libcosmic work. It doesn't seem critical.
- The way this is currently implemented seems like it results in e.g. having to go and set a profile for the HDMI device every time I plug it in, as it's set to "Off" when I unplug a device. The only exception is if I open the drop-down while it's plugged in, unplug it, and select the profile as if it was plugged in; this results in the drop-down being blank, but the profile is set to what I selected next time I plug it in. After re-plugging a couple of times, it's set to Off again on unplug, and then stays on Off again after I plug in again.
    - I'm assuming this kind of problem is related to why pavucontrol shows all profiles and just marks unplugged/unavailable ones as such, instead of hiding them.
        - On that topic, setting the profile to something that's `(unplugged) (unavailable)` in pavucontrol results in the COSMIC Settings drop-down being blank.
    - Not sure what the best way to solve this would be, but it seems like we need to remember the last profile in some form even if it becomes unavailable, rather than just setting it to off and forgetting whatever was previously selected.
        - Maybe if the selected profile becomes unavailable, we can allow it to remain in the list and stay selected until something else is manually chosen? I.e. filter to only show available *or* selected profiles, instead of just available ones. That would fix the view when selecting unavailable things from pavucontrol and prevent having to re-set the profile every time if something's getting plugged/unplugged a lot, while still retaining a cleaner look than pavucontrol.

I wouldn't be as concerned about remembering the last profile if the applet included profiles, but since UX has decided the applet won't show profiles, it seems like it'd be better for someone not to have to open up the Settings app and navigate into the Profiles page in order to get HDMI output as an option in the applet every time they plug into an HDMI device. But if this is too big of a challenge to tackle right now, let me know.

mmstick:
I'm able to add the unavailable device, but reloading the dropdown will require some changes in libcosmic that I'm working on. May be better to submit a followup PR for that.

mmstick:
Popups will refresh when selections change in libcosmic with https://github.com/pop-os/libcosmic/pull/1055

wash2 (approved):

jacobgkau:
I'm not sure the current solution to devices being plugged/unplugged is working right. I've had a couple of times where when I unplug HDMI, not only does the profile remain selected for the HDMI sound card (which is fine), but the Output device stays on the now-unplugged HDMI port (which is less desired). We probably want the Output device to change to something that's connected.

I also had one case on f24d602 where the HDMI device's profile was set to Off after unplugging and remained Off after plugging it in again, like it was before. But not every time.

jacobgkau:
Carl suggested we just automatically set a device to the first profile in the list when it used to have no profiles available (besides Off and Pro Audio) and a new profile becomes available. It seems like that would solve the usability issue, and ensure the output's at least available to select in the applet after plugging a device in. And would be simpler than trying to remember a history of profiles that were selected before things changed.

I am still seeing it often go to Off on unplug (and stay there on re-plug) on f24d602.

jacobgkau:
The HDMI plug behavior does seem to be behaving slightly better on 34cdc6b, but there are still some issues. For example, after about five plug/unplugs, I saw the profile remain on what it had been automatically set to on plug-in (`Digital Surround 7.1 (HDMI) Output`), instead of being changed to `Off` like it had the first few times I unplugged. pavucontrol shows the profile's `(unplugged) (unavailable)`, so the hardware knows it was unplugged.

jacobgkau:
It also seems that the profile being auto-selected when it was set to `Off` and I plug in HDMI is non-deterministic. It's been `Digital Surround 7.1` most times, but I just had it go to `Digital Surround 5.1` on one plug-in.

`Digital Surround 7.1 (HDMI)` is index 6. `Digital Surround 5.1 (HDMI)` is index 5. `Digital Stereo (HDMI)` is index 1, and is probably what I'd expect to be auto-selected (I'd think we'd either want that for the widest compatibility, or 7.1 to get the most channels possible). There are also a variety of `(HDMI 2)`, `(HDMI 3)`, etc. profiles shown in `pw-cli`, but cosmic-settings is correctly only showing the `(HDMI)` options when using the one HDMI port right now.

mmstick:
There may not be much we can do about what profile gets selected since we're competing with the pipewire daemon itself.

jacobgkau:
When testing HDMI plug/unplug on b4ca6d0, the first time I unplugged, the profile went back to Off (as expected with the behavior we're trying for now). The second and third times, it remained on the unplugged profile (which isn't the end of the world, but means the unavailable HDMI output is presented as an output device option and could remain selected). The fourth time, it was set to Off again.

@mmstick Just to clarify, is that also something the PipeWire daemon (and/or anything other than COSMIC Settings) is responsible for?

I did also just see one case when I plugged in where the profile flickered a whole bunch between the three (stereo, 5.1, and 7.1); it took a couple of seconds for it to settle down and pick the 5.1 to stay on. This spammed the OSD and volume adjustment sound while it was happening.

mmstick:
The daemon is responsible for switching profiles automatically, and that should be influenced based on plugged status and profile priority. But in all my testing it seems to be non-deterministic at times. Perhaps influenced by race conditions as device nodes are recreated.

We are attempting to work around the behaviors in cosmic-settings while the page is open, but that does have some limitations. In particular, there's no way to tell if a profile change was intentionally made by the user, or automatically by the daemon. Using the pipewire daemon's socket, we get messages from the daemon about active profile and route changes, but it doesn't indicate whether that was intentional by the user or an automatic selection by the daemon.

Currently, we're watching for changes after manually changing device profiles in the interface because the daemon tends to auto-select a random node for a different device during the time it takes for pipewire to recreate device nodes for the newly-selected profile. So for a brief moment a different device will get the default before cosmic-settings can re-select one of the newly-created nodes.

Now we have a conditional check for devices that are externally assigned to the Off profile while the page is open. The first profile with an available status will get selected when found. Though it's possible that pipewire may make that choice for us instead of switching to Off. So a patch to pipewire would be more appropriate for tweaking automatic profile behaviors.

> I did also just see one case when I plugged in where the profile flickered a whole bunch between the three (stereo, 5.1, and 7.1); it took a couple of seconds for it to settle down and pick the 5.1 to stay on. This spammed the OSD and volume adjustment sound while it was happening.

Checking. Sounds like a clash with one of our workarounds.

jacobgkau:
I see. Long-term, we probably want the auto-selection behavior to match whether Settings is open or not. Especially since Carl doesn't want normal users to ever have to open the Device Profiles page, there's no reason a user should have to be watching it, let alone think that watching it would influence the behavior.

Does PipeWire itself always select a profile (even if random) instead of Off when no profiles were available and one becomes available, or is that part of the workarounds in COSMIC Settings? Would it make sense to move any of these hotplug workarounds to cosmic-settings-daemon?

mmstick:
> Does PipeWire itself always select a profile (even if random) instead of Off when no profiles were available and one becomes available

Pipewire should automatically select a profile with the highest priority on plug if the Off profile was not saved. Setting a device's profile to Off in cosmic-settings—and maybe pavucontrol too— saves it, so it may not auto-select in that case? The same may be true if an unplugged profile was set with the saved property. The best way to test would be to watch pipewire's status without cosmic-settings or the audio applet running in the background.

> Would it make sense to move any of these hotplug workarounds to cosmic-settings-daemon?

Patching pipewire directly would be more reliable, but since we can't ensure that other distributions have the same patches, automatic device and profile selections would be better off in the cosmic-settings-daemon instead of cosmic-settings. Leaving cosmic-settings to only worry about applying changes instead of reacting to them.

The only workaround cosmic-settings may need itself is to re-select a default device after we've changed its profile in the device profiles page. Though perhaps we could instead set that in a config with cosmic-config so that cosmic-settings daemon can also track settings set intentionally by cosmic-settings.

jacobgkau:
On 2d09539, I'm seeing that if I have the main Sound page of Settings open, when I plug in HDMI, a profile often doesn't get selected for it and I end up not being able to select it as an output device-- *until* I click into Device Profiles, at which point it immediately selects a profile instead of Off, and auto-selects the device as an output. This would be even more confusing to a user than having to manually turn it on, since it appears visually like it's already on when entering the Device Profiles page, even though it wasn't until that page was entered.

Likewise, if I'm sitting in the Device Profiles page, I'm seeing the profile stay Off until I exit the page-- at which point a profile's selected, and it's shown if I enter the page again.

With cosmic-settings not running and the Sound applet removed from the panel, pavucontrol seems to be getting the profile right on plug/unplug every time *after* setting it at least once in pavucontrol. (Before I set it in pavucontrol, it turned off on unplug and stayed off on re-plug.)

mmstick:
`env RUST_LOG=sound=debug cargo run` will give debug logs for just the sound page now.

mmstick:
Okay I've reworked all of the logic for auto-selecting devices when the cosmic-settings page is open

- 6e908b3...ba0c7c5 Removed all profile / device auto-select logic 
- eeec525 Auto-select device if active device's profile is changed internally (on profile dropdown select) 
- 84ff320 Auto-select device if active device's profile is changed externally (such as by hotplug)

jacobgkau:
Can we rebase this again? I'm having problems with it sometimes crashing after I remove the COSMIC config folder, I'm assuming maybe because of 41e90bde337d03b269ab67142382d24faeb96255. The Panel and Dock pages are also sometimes broken (I'm assuming because of missing the same commit), which makes it difficult to remove the Sound applet for testing.

Also, I'm seeing some instances where e.g. after unplugging HDMI while on the `Device profiles` page, the drop-down still has `Digital Stereo (HDMI) Output` selected. `pw-cli e 50 Profile` shows that it's actually been set to Off, but COSMIC Settings doesn't see that until I go back to the main Sound page-- I see a flicker in the output devices that suggests the page figures it out then, and then it's `Off` if I navigate back into `Device profiles`. So I think the refresh/update isn't working there.

jacobgkau:
After looking at WirePlumber's debug logs, I think Pop!_OS is configured to have WirePlumber handle auto-selecting/remembering device profiles, not the PipeWire daemon itself. (PipeWire has an ALSA option for it, but the docs note it's usually disabled so the session manager can handle it instead.)

I think it's working correctly. But again, on a44fc56, I'm seeing COSMIC Settings not update to reflect the profile change when I unplug HDMI, until I exit and re-enter the `Device profiles` page. It does update to go from `Off` to `Digital Stereo (HDMI) Output` when I plug in; it's only when I unplug that it's not updating to `Off` until I manually re-enter the page.

jacobgkau:
I'm tearing my hair out trying to test at this point. I'm getting the following crash on this branch very often after trying to reset the config files:

```
system76@pop-os:~/cosmic-settings$ RUST_BACKTRACE=full COLORBT_SHOW_HIDDEN=1 cosmic-settings
  ERROR  shortcuts custom config error: GetKey("custom", Os { code: 2, kind: NotFound, message: "No such file or directory" })
    at /build/cosmic-settings-eJXdZR/cosmic-settings-0.1.0~1763999493~24.04~6f4c755/vendor/cosmic-settings-config/src/shortcuts/mod.rs:43 on main

  ERROR  Failed to load Panel system config., err: GetKey("padding_overlap", Os { code: 2, kind: NotFound, message: "No such file or directory" })
    at cosmic-settings/src/pages/desktop/panel/mod.rs:90 on main

  ERROR  Error when loading Panel container config., err: GetKey("padding_overlap", Os { code: 2, kind: NotFound, message: "No such file or directory" })
    at cosmic-settings/src/pages/desktop/panel/inner.rs:71 on main

  ERROR  Error when loading Panel container config., err: GetKey("padding_overlap", Os { code: 2, kind: NotFound, message: "No such file or directory" })
    at cosmic-settings/src/pages/desktop/panel/inner.rs:71 on main

  ERROR  Failed to load Dock system config., err: GetKey("padding_overlap", Os { code: 2, kind: NotFound, message: "No such file or directory" })
    at cosmic-settings/src/pages/desktop/dock/mod.rs:139 on main

  ERROR  Error when loading Panel container config., err: GetKey("padding_overlap", Os { code: 2, kind: NotFound, message: "No such file or directory" })
    at cosmic-settings/src/pages/desktop/panel/inner.rs:71 on main

  ERROR  Error when loading Panel container config., err: GetKey("padding_overlap", Os { code: 2, kind: NotFound, message: "No such file or directory" })
    at cosmic-settings/src/pages/desktop/panel/inner.rs:71 on main

  ERROR  shortcuts custom config error: GetKey("custom", Os { code: 2, kind: NotFound, message: "No such file or directory" })
    at /build/cosmic-settings-eJXdZR/cosmic-settings-0.1.0~1763999493~24.04~6f4c755/vendor/cosmic-settings-config/src/shortcuts/mod.rs:43 on main

The application panicked (crashed).
Message:  assertion failed: self.is_char_boundary(idx)
Location: cosmic-settings/src/pages/sound/mod.rs:484

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   1: __libc_start_main<unknown>
      at <unknown source file>:<unknown line>

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
Aborted (core dumped)
system76@pop-os:~/cosmic-settings$
```

I don't think any of the `ERROR` lines are relevant (they also happen on master), but do you know why the `assertion failed` thing would be happening? I can downgrade to master, Settings opens fine, then I upgrade back to this branch and it won't launch in this state.

jacobgkau:
Just now, I saw Settings start working after I suspended for a while and then resumed. But after removing the config files and rebooting again, it stopped working again.

Here is a more complete output doing a `cargo run` with the debug log: [sound-crash.txt](https://github.com/user-attachments/files/23732321/sound-crash.txt)

mmstick:
I think that code isn't needed anymore, so we can remove it.

mmstick:
The applet is pointing to an older commit, so I'll update it.

mmstick:
The applet PR has been rebased. The current behavior is to only save when manually saving a profile.

jacobgkau (approved):
Working as expected on a desktop as well.

Issue #1318, opened 2025-08-18
Bluetooth speaker shows "Disconnecting" for most of connection process


jacobgkau:
Version: fbe004b

When connecting to a Roverbeats T3 speaker from a laptop (bonw16, serw14, or gaze20), after clicking the speaker in the "Nearby Devices" list, the speaker moved to "Previously Connected Devices" and then showed its status as "Disconnecting" for at least several seconds:

<img width="3840" height="2400" alt="Image" src="https://github.com/user-attachments/assets/a2e07de7-f68d-49c8-827c-c792e46920da" />

Eventually, "Disconnecting" was replaced by the Connect button, and it then switched to "Connected" without me having to click it (which all took a fraction of the time "Disconnecting" was displayed). This makes the process look confusing; whatever status is being reported as "Disconnected", it should probably say "Connecting or Disconnecting" if we can't differentiate between the two (or just "Connecting", since disconnecting seems to be nearly instantaneous-- it does show "Disconnecting" for a split second, but far less than "Disconnecting" was shown while connecting).

brandonbeaumia:
Yes, I can reproduce this on multiple bluetooth controllers as well. Definitely hard to figure out what's going on if you're not staring at the controller lights (assuming it even has them).

PR #1501, opened 2025-10-20
fix: use rfkill when toggling bluetooth

### LINKED ISSUE #1498: Settings can't turn Bluetooth on if it was last turned off by the applet
cosmic-settings: 086f501
cosmic-applets: ae4a2a7

If I turn off Bluetooth in the applet, then turning it back on in Settings doesn't work (no devices are ever shown). I have to turn it back on in the applet (then sometimes also toggle it again in Settings) before it will work again.

https://github.com/user-attachments/assets/ea75cb32-3dd8-4dca-afde-21b9cd6d8b7e

If I just turn it off from Settings, then I can turn it back on in Settings and it works.

wash2:
fixes #1498 & #1318

### DIFF
--- a/subscriptions/bluetooth/src/device.rs
+++ b/subscriptions/bluetooth/src/device.rs
@@ -111,7 +111,7 @@ impl Device {
                     }
                 }
                 DeviceUpdate::Paired(paired) => {
-                    self.enabled = Active::Disabling;
+                    self.enabled = Active::Enabling;
                     self.paired = paired;
                 }
                 DeviceUpdate::Icon(icon) => self.icon = icon,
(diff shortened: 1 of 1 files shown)


garrettjwilke (dismissed):

mmstick (dismissed):

subscriptions/bluetooth/src/device.rs
```diff
@@ -255,6 +255,8 @@ pub async fn connect_device(connection: zbus::Connection, device_path: OwnedObje
         }
     }
 
+    dbg!("failed to connect after retries");
+    dbg!("failed to connect after retries");
```
jacobgkau: Are there supposed to be two of these log lines?

mmstick (approved):

PR #1511, opened 2025-10-27
[Applets] Always show placeholder on empty segments

git-f0x:
This makes it clearer that applets can be placed in those segments before a drag is started, and prevents things moving around when the drag is started. Design discussed with UX.
Screenshot_2025-10-29_17-27-29

I changed the text to say "Place" instead of "Drop", which makes it fit both regular and DnD, and matches COSMIC Store verbiage (deleting the existing translation entries is probably the best way since Weblate is there).

Also addressed most clippy lints, but without touching anything sound-related (to avoid potential conflicts with the existing sound redesign PR).


jacobgkau (approved):
There's still a slight flicker when you start dragging an applet in the list, as it disappears from the list before its drop placeholder is re-added (sometimes in a different spot, depending on the mouse cursor position). But this does reduce movement during dragging (and looks better).

The applet rearranging pages for the panel and the dock are still working as expected, with the new change. I've also spot-checked all the pages with lint fixes:

- [X] Default Applications
- [X] X11 Applications Compatibility
- [X] Startup Applications
- [X] Bluetooth
- [X] Appearance (font context drawer)
- [X] Appearance (icon theme context drawer)
- [X] Appearance (density settings)
- [X] Panel (main page, display setting)
- [X] Wallpaper
- [X] Window management (Super key binding)
- [X] Displays (arrangement)
- [X] Displays (main page, resolution & enable/disable)
- [X] Keyboard (layout)
- [X] Keyboard Shortcuts
- [X] Wi-Fi
- [X] Power
- [X] Users
- [X] Date & Time (region selection)

git-f0x:
> There's still a slight flicker when you start dragging an applet in the list

Yeah, I've noticed that, but I'm not sure how that could be prevented currently.
Also, GitHub seems to want this to be rebased (or at least says the branch is out-of-date). Not sure if that's important, since there aren't conflicts. Should I do it (will likely dismiss the review)?

jacobgkau:
> Also, GitHub seems to want this to be rebased (or at least says the branch is out-of-date). Not sure if that's important, since there aren't conflicts. Should I do it (will likely dismiss the review)?

I checked on that before I built the branch for testing. The only commit not in this branch yet is a8d8e14413d0488963f6485c89126a203ebdb651, which was just a translations update. I think this repo's set up such that we'll be able to rebase & merge in one go once we have an engineering approval. I can approve again for QA if you want to rebase it yourself, though.

wash2 (approved):

PR #1512, opened 2025-10-27
Fix compiler warnings

ids1024:
I was also going to fix runtime errors here, but it seems https://github.com/pop-os/cosmic-panel/pull/490 fixed that instead.


git-f0x:
https://github.com/pop-os/cosmic-settings/pull/1511 does the same 😅

PR #1523, opened 2025-11-01
Setting to format date and time with strftime

joshuamegnauth54:
Depends on: pop-os/cosmic-applets#1129

---

See the original MR for screenshots and more info. Draft until I test and until the original MR is accepted.


mmstick (approved):

leviport (changes requested):
Customizing the format is working nicely, but making customizations in the new settings input is presenting a usability problem.

The value is updated right away, so if I add some characters that aren't `%` codes, they show up immediately. However, when I try adding `%` codes, just typing the `%` sign causes the time applet to panic and disappear from the panel, printing this in journal logs:

```
Nov 10 13:52:48 pop-os cosmic-panel[10008]: com.system76.CosmicAppletTime:
Nov 10 13:52:48 pop-os cosmic-panel[10008]: com.system76.CosmicAppletTime: thread 'main' panicked at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/alloc/src/string.rs:2831:14:
Nov 10 13:52:48 pop-os cosmic-panel[10008]: com.system76.CosmicAppletTime: a Display implementation returned an error unexpectedly: Error
Nov 10 13:52:48 pop-os cosmic-panel[10008]: com.system76.CosmicAppletTime: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Nov 10 13:52:49 pop-os cosmic-panel[10008]: com.system76.CosmicAppletTime: exited without error
Nov 10 13:52:49 pop-os cosmic-panel[10008]: com.system76.CosmicAppletTime: exited with code 137
```

It happens so fast that I can't even get the next letter in there after the % sign. If I type out the strings I want into a text editor, then copy/paste them in, it works fine. I just can't type them manually.

After that crash, I can restore the time applet is to `killall cosmic-panel` to restart the panel and all its applets (as long as there are no invalid `%` codes in the format box), or I can log out and back in again. Just clearing or correcting the format box does not restart the time applet.

leviport:
I also noticed that custom formats don't play very nicely with vertical panels:

Screenshot_2025-11-10_15-07-55

This was probably more of an issue with https://github.com/pop-os/cosmic-applets/pull/1129, but exposing it in cosmic-settings makes it seem like it should play more nicely with other possible cosmic-settings customizations.

joshuamegnauth54:
@leviport Weirdly enough I didn't even notice this panic. My time applet disappears if I input `%` but shows up again after I complete the specifier. I assumed that it disappeared because the specifier was empty not because of a panic. However, I checked the logs and it's definitely panicking.

I'll work on this soon. My guess is that it's an edge case where an empty string from an incomplete initial specifier is causing a panic in the applet widget.

dyrland:
I'm excited for this to come to fruition! (I'm waiting for it to be exposed in `cosmic-settings`; have not created a file called `strftime`.)

joshuamegnauth54:
I fixed it, but the fix is here: pop-os/cosmic-applets#1158 :grin: Nothing needs to change in this PR.

@dyrland I'm glad you're excited! I have time to work on it this weekend so I should have everything finished soon.

@leviport I'm not sure what to do with the vertical view. :thinking: It may be better if I disable strftime for the vertical view because it's hard to make it look nice. If you would prefer if I disable it, let me know and I'll do it tomorrow in pop-os/cosmic-applets#1158.

dyrland:
I think you leave it on for vertical...people should just use "Mon" not "Monday" and "Nov" not "November", right? People can make their own decisions?

I mean, I love that COSMIC automagically makes it pretty, but if I want to go custom, I don't expect the magic to keep working.

joshuamegnauth54:
I think that's reasonable too. The vertical view would require more effort to look nice, but in the spirit of customization I'm sure users would undertake that challenge.

leviport:
We probably need to get @pop-os/ux 's  input on vertical panels. I share the sentiment that enabling deeper customization can grant the ability to make something weird/broken looking, so as a user, I would expect to either leave it default or play with it until I found variables that fit nicely in a vertical panel. I'll re-test with https://github.com/pop-os/cosmic-applets/pull/1158 to make sure the panics are gone and let UX weigh in on vertical panels.

leviport (approved):
With https://github.com/pop-os/cosmic-applets/pull/1158, this is working very nicely. We should still get UX's feedback before merging this, but from a functionality standpoint, I think it is good to go.

maria-komarova:
Thanks for working on it. The options need further thinking and design before merging could be considered. As it is currently it is not easy to understand for non-power users. Even I had to search for strftime to understand what it is. And I don't see a quick way to explain it to people. Aside from that the formats will need to be readable and look good with the vertical panel, this is part of making the user experience good on multiple levels.

leviport:
Maybe it would be better to expose the option somewhere like [COSMIC Tweaks?](https://github.com/cosmic-utils/tweaks)

dyrland:
I was going to say, this should definitely be exposed somewhere. COSMIC and Linux are all about user controls, right?

I wonder what UX would think about a little helper link that brings someone to strftime? Like one of those question mark hyperlink things (which doesn't seem to match the current COSMIC UI, so this feels like a non-starter idea that might lead to a good idea?)

At the very least, could this be exposed in the FAQ for power-users so they don't have to dig through Git?

mmstick:
@dyrland First party applets must be responsive to horizontal and vertical panel layouts. If a customization is provided that breaks that expectation, it's a bug. Ideally there should be some way of automatically adjusting the format to fit in both layouts.

COSMIC has a lot of configurable options, but they're always exposed in COSMIC Settings after careful consideration about their design, behavior, and side effects. COSMIC Tweaks would be more appropriate for experimental features that don't yet have an approved design for COSMIC Settings.

joshuamegnauth54:
Fair enough. Should I port this over to COSMIC Tweaks? :thinking:

BenjaminBrienen:
I would really like to not have to edit a hidden text file to customize the time format string...

PR #1526, opened 2025-11-03
Makes it possible to disable the panel

### LINKED ISSUE #1525: Make it possible to disable the Panel
Make it possible to disable the Panel.  
The dock can be disabled already so it would feel more consistent if it would be possible to disable both.   

Settings -> Desktop -> Dock -> Dock Toggle
Settings -> Desktop -> Panel -> Add (Panel Toggle)

Nianmi:
Closes #1525 
Adds a Panel toggle to the panel page. Now the Panel can be toggled like the dock. 
Fells more consistent that dock and panel have the same settings.

### DIFF
--- a/cosmic-settings/src/pages/desktop/dock/mod.rs
+++ b/cosmic-settings/src/pages/desktop/dock/mod.rs
@@ -1,13 +1,9 @@
 use std::collections::HashMap;
 
-use cosmic::{
-    Task,
-    cosmic_config::{ConfigSet, CosmicConfigEntry},
-};
+use cosmic::{Task, cosmic_config::CosmicConfigEntry};
 use cosmic_panel_config::{CosmicPanelConfig, CosmicPanelContainerConfig};
 use cosmic_settings_page::{self as page, Section, section};
 use slotmap::SlotMap;
-use tracing::error;
 
 use crate::pages::desktop::panel::inner::{
     add_panel, behavior_and_position, configuration, enable, reset_button, style,


jackpot51:
There is concern that turning off the panel would leave users in a state that they cannot recover from easily.

Nianmi:
Aaa i see your point. I think the risk of this happening is minimal.  
On the other hand i'm certain there are users out there who wish to neither have a panel nor a dock.

Nianmi:
how about we make sure that the reset to default button also resets the enable in Dock and Panel? So far this setting is ignored.

jacobgkau:
While I personally agree the risk of this being a problem is small-- truly getting locked out from Settings to re-enable the panel would require re-binding the Super key to Workspaces, remapping the Super-/, Super-A, and Super-T shortcuts to other things without mapping replacements, and killing cosmic-files-applet after breaking its normal ability to auto-restart-- internal discussion with the UX team sounds like we're going to hold off on it, at least until COSMIC is more established and we've got more user testing under our belts.

The good news is that this option is already present in the community [COSMIC Tweaks](https://github.com/cosmic-utils/tweaks) app, so for anyone who feels auto-hide isn't good enough, that's an easy way to accomplish turning the panel off (as an alternative to just editing the config file).

Nianmi:
Sounds good to me. I will leave the issue open so we can see if anybody else would like this option.  
No need for cosmic tweaks i just delete all the applets inside of the panel. As soon as it is empty the panel is no longer displayed. This workaround might be considered a bug if you want to force the panel open at all time.  

This was the reason for the PR there is already a way to hide the panel it involved just extra steps with deleting all applets. While disabling the dock was straight forward. I thought this inconsistency was weird.  

Anyways i don't think that there are than many people out there like me which don't need a dock or pannel.

PR #1541, opened 2025-11-05
adjust oversized scroll area for display arrangement

### LINKED ISSUE #817: Pointless display scrollbar can still sometimes appear
Whilst it appears that the issue with too much space being present in the display settings has been fixed, it appears to still be able to show up under certain conditions

Upon first loading the settings, the scroll bar is still present

screenshot-2024-12-12-10-41-05

Ignore the red border the window is fullscreened but I had to go back in my stream to grab the screenshot

After I changed a setting I guess the UI refreshed and now it is gone as it should be

screenshot-2024-12-12-10-41-51

FreddyFunk:
The bounding box for the scroll area in the display arrangement configuration page was way to wide, which looks kind of broken in most situations. This properly positions and resizes the widths. 

Should fix #817 as well.

before:

https://github.com/user-attachments/assets/906df537-4d71-43be-a51c-60a8bf1c6451

after:

https://github.com/user-attachments/assets/26abfc72-60d2-471d-961d-829b0e407bb4

### DIFF
--- a/cosmic-settings/src/pages/display/arrangement.rs
+++ b/cosmic-settings/src/pages/display/arrangement.rs
@@ -141,11 +141,11 @@ impl<Message: Clone> Widget<Message, cosmic::Theme, Renderer> for Arrangement<'_
         let state = tree.state.downcast_mut::<State>();
 
         state.max_dimensions = (
-            max_dimensions.0 as f32 / UNIT_PIXELS,
+            max_dimensions.0 as f32 / UNIT_PIXELS / 2.0,
             total_height as f32 / UNIT_PIXELS,
         );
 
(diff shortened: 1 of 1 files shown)


FreddyFunk:
This issue appears to be that as soon as the iced scrollable area has no scrollbar (e.g. does not need to be scrollable), the underlying custom draw function gets called at a different time, effectively changing the z-order, so that the custom draw function gets called later. This results in the overlapping of the "displays" with other UI elements. Alternatively one could rewrite the custom draw function, but this would be a bigger task, since zooming, padding for comfortable re-arranging of displays and moving focus would need to be thought of.

I will look for another task and mark this PR as draft. If anyone would like to pick up this PR, feel free to do so.

PR #1542, opened 2025-11-05
syntax and light copy changes

maria-komarova:
Changed title case to sentence case, remove unnecessary punctuation, change a few words in the copy.


therealmate:
Do the .desktop files need to mirror these changes?

PR #1544, opened 2025-11-06
[Keyboard] Remove extra padding from special character and numlock drawers

git-f0x:
Removes the extra padding around the content in those drawers, which doesn't seem to be there in the designs, and makes these drawers stand out.
Also added a (slightly modified for better flow) description from the designs for the Compose key.

Reverts the color picker drawer title changes, since the updated context drawer design makes it look ok.


wash2 (approved):

git-f0x:
Had to force push because of a conflict (in Cargo.lock).

PR #1545, opened 2025-11-06
[Display] Make "Don't Mirror" disable display-projektion

QuantenToast:
Without this pressing “Don't mirror” on a display that is projected onto another does nothing, instead of stopping the projection.

### DIFF
--- a/cosmic-settings/src/pages/display/mod.rs
+++ b/cosmic-settings/src/pages/display/mod.rs
@@ -16,6 +16,7 @@ use cosmic_randr_shell::{
     AdaptiveSyncAvailability, AdaptiveSyncState, List, Output, OutputKey, Transform,
 };
 use cosmic_settings_page::{self as page, Section, section};
+use indexmap::Equivalent;
 use slab::Slab;
 use slotmap::{Key, SecondaryMap, SlotMap};
 use std::sync::atomic::{AtomicBool, Ordering};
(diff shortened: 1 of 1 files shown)


cosmic-settings/src/pages/display/mod.rs
```diff
@@ -546,7 +547,23 @@ impl Page {
...
 
             Message::Mirroring(mirroring) => match mirroring {
-                Mirroring::Disable => return self.toggle_display(true),
+                Mirroring::Disable => {
+                    for k in self.mirror_map.keys() {
+                        if k.equivalent(&self.active_display) {
+                            return self.toggle_display(true);
+                        } 
+
+                        if let Some(v) = self.mirror_map.get(k) && v.equivalent(&self.active_display) {
```
mmstick: let expressions are unstable in Rust 1.85.0, so I'll need to patch this.

jacobgkau (approved):
Before, clicking `Don't mirror` on a display mirroring a different display did work; but clicking `Don't mirror` on a display being projected to a different display did nothing.

Now, clicking `Don't mirror` works to disable mirroring whether it's from the mirrored screen or the projected screen.

Regression testing passed for the Displays page (display rearrangement & all other settings still work). Also set up a three-display setup to ensure the mirror/project options still work as expected.

mmstick (approved):

PR #1551, opened 2025-11-07
re-design date & time page and language & region

FreddyFunk:
This is a re-design for the setting pages to configure date, time, language and region.

This is going to be a series of pull request targeting those areas. This pull request lays the ground work to easily expand further changes in to separate, independently testable and merge-able changes. Things I will work on once this PR is merged:

Language & region:
- Add option to configure Calendar (Gregorian, Japanese, Buddhist) to be different than the regions default
- Add option to configure Temperature (Celsius, Fahrenheit) to be different than the regions default
- Add option to configure Measurement System  (Metric, Imperial (US), Imperial (UK)) to be different than the regions default
- Add option to configure First Day of Week (and all week days allowed) to be different than the regions default
- Add option to configure date format to be different than the regions default
- Add option to configure number format to be different than the regions default
- Prettier formatting preview

Date & time:
- Add toggle option to set time zone automatically based on location

based on #1549

### before:

Screenshot_2025-11-07_23-47-42
Screenshot_2025-11-07_23-47-51
Screenshot_2025-11-07_23-47-57
Screenshot_2025-11-07_23-48-09

### after:
Screenshot_2025-11-07_23-50-55

https://github.com/user-attachments/assets/b34f4cb6-1715-44d0-9a3c-446c5d1b11e5

Screenshot_2025-11-07_23-51-28


jacobgkau:
This is going to need UX review. I mentioned the [Pop!_OS Mattermost chat server](https://chat.pop-os.org) in another one of your PRs today; you are highly encouraged to join that if you're going to be wanting to submit large visible changes like this, just to avoid spending extra effort going directions that may not be desired.

The additional options are hopefully okay, but I'm not sure we'll want to move these from being sub-pages to top-level pages. I also don't think we're going to want to move the time formatting options to the Panel section, because you could put a time applet in the Dock as well.

Issue #885, opened 2025-01-11
[Feature Request] Add number indicator on display configuration


BrodieRobertson:
Some desktops provide the ability to display a number on your displays in the display configuration so that it's easy to know which of the displays in the layout correspond to the physical monitors. Especially useful if you have multiple of the exact same monitor

PR #1554, opened 2025-11-09
trigger cosmic-osd display identifier OSD on each screen to ease display arrangement configurations for users

FreddyFunk:
implements #885

requires https://github.com/pop-os/cosmic-osd/pull/151

preview:

https://github.com/user-attachments/assets/2b008908-1b79-4662-9b92-fa876353c445

### DIFF
--- a/cosmic-settings/src/pages/display/mod.rs
+++ b/cosmic-settings/src/pages/display/mod.rs
@@ -151,7 +151,7 @@ pub struct Page {
     active_display: OutputKey,
     randr_handle: Option<(oneshot::Sender<()>, cosmic::iced::task::Handle)>,
     hotplug_handle: Option<(oneshot::Sender<()>, cosmic::iced::task::Handle)>,
-    display_identifier_handle: Option<(oneshot::Sender<()>, ())>,
+    display_identifier_handle: Option<(oneshot::Sender<()>, cosmic::iced::task::Handle)>,
     display_identifier_guard: Option<DisplayIdentifierGuard>,
     config: Config,
     cache: ViewCache,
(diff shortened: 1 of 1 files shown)


maria-komarova:
Thanks for working on it. There was a previous discussion about it and a [mockup](https://www.figma.com/design/Z6Z9Ox6WNkiiM8gcAkcElS/COSMIC-Settings-First-Release---Developer-Handoff?node-id=33602-14967&t=Hkf3tn8x6G24OfwZ-4). The idea is to always show the numbers when there are 2+ displays connected. We've seen cases where people struggled to identify which display is which while trying to arrange them so having the numbers always present while Displays page in Settings is focused would be helpful. You can see the placement and the styling of the OSD on the mockup. Accent color background is used to make the OSD more immediately apparent.

wash2:
I think that the way this is implemented is mostly fine, but there should probably be some sort of timeout for the osd, so that if settings fails to dismiss it properly, it disappears on its own after a few seconds. For example, when testing if, if I navigate to the displays page and close the settings app, the osd does not disappear unless I navigate away from the page.

@pop-os/ux Do you think it could appear on first entering the page, then fade away after a few seconds, and reappear when a display is hovered?

maria-komarova:
> Do you think it could appear on first entering the page, then fade away after a few seconds, and reappear when a display is hovered?

I was wondering about that but we've discussed it with @WatchMkr and thought it best to have the numbers show as long as the Displays page is focused.

FreddyFunk:
> I think that the way this is implemented is mostly fine, but there should probably be some sort of timeout for the osd, so that if settings fails to dismiss it properly, it disappears on its own after a few seconds. For example, when testing if, if I navigate to the displays page and close the settings app, the osd does not disappear unless I navigate away from the page.

This should only happen, if you forcefully kill the cosmic-settings process. When you gracefully close the settings app via the "x" button, then the OSD should disappear as well. Nonetheless, during the development I have implemented two different proof of concepts to deal with a cosmic-settings app crash/kill and still be able to hide the display identifier OSD.

1. Add a listener to the display identifier OSD that checks if the process ID that has triggered the event is still running. If not, the OSD close on their own.
2. The OSD closes after a 2 second timer, but sending the identify displays d-bus message resets this timer for all display identifier OSDs.

Option 1 feels a bit hacky, option 2 feels like unnecessary message spam on the d-bus (but manageable, since its a 1-to-n connection from sender to subscriber). I did not feel comfortable with either option so I did not included them. Feel free to let me know what you think would be the best option.

> @pop-os/ux Do you think it could appear on first entering the page, then fade away after a few seconds, and reappear when a display is hovered?

I can implement a fade out animation as well. The more specific you are with those details, timings and so on, the better.

FreddyFunk:
Here is a video of what the current develop state in this pull request looks like:

https://github.com/user-attachments/assets/1369086c-09c5-4cee-9352-6eff00048555

wash2:
> The OSD closes after a 2 second timer, but sending the identify displays d-bus message resets this timer for all display identifier OSDs.

This sounds fine to me for now. Maybe even with a 1 second timer.

> This should only happen, if you forcefully kill the cosmic-settings process. 

It happened when I closed settings using the `Super + Q` shortcut

FreddyFunk:
Alright, then I will do it that way.

FreddyFunk:
addressed the feedback

wash2 (approved):

FreddyFunk:
@jacobgkau  oh wow, good catch! I haven't thought about this scenario with the display hot-plugging. I have pushed fixes for that and the visual changes to the OSD pull request.

FreddyFunk:
I was able to reproduce this error by rapidly moving opening the display page, then any other settings page and then the display page again in a loop to reliably recreate the bug. The debugging took me down a rabbit hole and I found a bug in the pop-os iced fork used by libcosmic used by cosmic-osd. This bug caused a Wayland layer-shell protocol violation: "must ack the initial configure before attaching buffer" which occurres when rapidly creating and destroying layer surfaces.

I have opened a pull request for the fix here: https://github.com/pop-os/iced/pull/250

Once this is merged, libcosmic needs to be updated with the new iced version and then I can update the libcosmic version in the cosmic-osd pull request that is linked to this pull request here.

FreddyFunk:
I tweaked the visual size of the OSDs again. Hopefully I managed to get this nailed down now.

FreddyFunk:
I updated libcosmic on the OSD pull request which includes the previously mentioned iced fix. This can now be tested again.

FreddyFunk:
I was able to reproduce the bug and pushed a fix for that.

jacobgkau (commented):
On c58dd10, the bug I described seems to be partially fixed when testing with the Bluetooth page, but it's still happening when I test with the Desktop page. (E.g. if Displays is open, attempting to go to the Desktop page via the Launcher flickers and stays on the Desktop page.)

It also still happens sometimes with the Bluetooth page if I'm using the Launcher from the same workspace Displays is open on.

jacobgkau:
Actually, what I'm seeing now might not be a regression. I'm seeing similar behavior with other pages on another branch of cosmic-settings now.

jacobgkau:
For posterity, this was reverted in https://github.com/pop-os/cosmic-settings/pull/1605 because it caused https://github.com/pop-os/cosmic-epoch/issues/2579 (broke display drag/drop). Info from Michael:

> cosmic-randr always sees two dispatched EventQueue events when dispatching so it keeps reloading.

mmstick:
Yes, we use `cosmic-randr::Context` and its `async_dispatch` method to watch for `ManagerDone` events. When that message is received, we reload the page's display data by calling `cosmic-randr list --kdl`. And since cosmic-osd was being asked to run `cosmic-randr list --kdl` every 0.5 seconds, that would cause `ManagerDone` events to be seen each time. Creating an endless loop of reloading the page. Need to find a different mechanism. And cosmic-osd shouldn't need to run `cosmic-randr list --kdl` that frequently either.

PR #1565, opened 2025-11-12
Refactor system/about page and improve GPU name

### LINKED ISSUE #892: Report AMD GPUs correctly
I've seen some applications like my terminal fetch application report the AMD GPU correctly

screenshot-2025-01-11-12-23-20

Under the COSMIC about sections it instead shows it in this incredibly verbose way where the correct GPU is reported but it also shows a bunch of others.
screenshot-2025-01-11-12-26-56

I've seen a lot of other applications report my GPU in the same way so it seems like a common issue with these types of tools

FreddyFunk:
closes #892

also refactors the about page by:
- simplified the information gathering logic for the about page
- removing 4 build time dependencies (bumpalo, byte-unit, concat-in-place, const_format) resulting in a reduction in the cargo try by 14 (2177 -> 2163)
- removing runtime dependencies (lscpi)
- gathering GPU names with wgpu instead of CLI tools, since wgpu is already used by iced anyway

before:

image

after:

image

### DIFF
--- a/cosmic-settings/src/pages/system/info.rs
+++ b/cosmic-settings/src/pages/system/info.rs
@@ -249,7 +249,10 @@ fn processor_name() -> String {
     if let Some(cpuinfo) = read_to_string("/proc/cpuinfo") {
         for line in cpuinfo.lines() {
             if let Some(info) = line.strip_prefix("model name") {
-                if let Some(info) = info.trim_start().strip_prefix(':') {
+                if let Some(mut info) = info.trim_start().strip_prefix(':') {
+                    if let Some(cpu_with) = info.find(" w/ ") {
+                        info = &info[..cpu_with];
+                    }
                     return info.trim().to_string();
                 }
                 break;


mmstick:
How does the performance compare for loading and refreshing the data? This page is somewhat performance-sensitive due to all the buffering and parsing behind the scenes. Hence the use of bumpalo as a reusable buffer space for pooling allocations while collecting data.

FreddyFunk:
I do not notice any significant changes in performance.

before:

https://github.com/user-attachments/assets/e9dccc0a-fe36-44ff-bd39-7ac52d0cd285



after:


https://github.com/user-attachments/assets/880df299-cce5-48bc-8428-9b42ca8b7ddc

garrettjwilke:
not sure if the changes will be important, but Intel GPU does not show the "full name" or type of GPU, it simply says "Intel"

this is `cosmic-settings` before this change:
IMG_3227

and here it is after:
IMG_3226

here is another system with both intel and nvidia before:
IMG_3228

and after:
IMG_3229

FreddyFunk:
@garrettjwilke 
> not sure if the changes will be important, but Intel GPU does not show the "full name" or type of GPU, it simply says "Intel"

Thanks for the hint. So this is definitely also an improvement for NVIDIA GPUs as well. 

For the Intel issue... I currently have no device with an Intel integrated GPU available, but this looks like a quirk of the Intel GPU driver.  If you can install GPU Viewer and post a screenshot of the Vulkan page (with open drop down in the top middle "Available Devices") and the OpenGL page, I might be able to fix that.

https://flathub.org/en/apps/io.github.arunsivaramanneo.GPUViewer

jacobgkau:
Here are the screenshots you requested on a darp11, which has the behavior that Garrett described:

Screenshot_2025-11-13_14-50-47
Screenshot_2025-11-13_14-51-05

jacobgkau:
It looks like the "ARL" part of some of those strings signifies that it's Arc graphics on Arrow Lake, so that would be a useful identifier for that one.

FreddyFunk:
The Intel GPU drivers do not properly report their name... I added a quirk for Intel GPUs, based on this documentation that I found:
https://www.intel.com/content/www/us/en/support/articles/000005520/graphics.html

I tried my best to get the formatting right with dummy test data, but someone with a Intel GPU system needs to test that. If it does not show the name of the Intel GPU as you expect, please share the output of the following command:
```bash
lspci -k | grep -EA3 'VGA|3D|Display'
```

I also fixed a bug that would not show all GPUs, in case a user has multiple GPUs of the exact same model installed.

FreddyFunk:
I got my hands on a laptop with an Intel iGPU and NVIDIA GPU was able to validate and fix the following:

- Intel iGPU Name formatting works now
- NVIDIA Optimus GPUs get now detected, which they do not in the current master build

before:
before


after:
after

jacobgkau (commented):
This is a before/after on a darp8. Before:

Screenshot_2025-11-24_14-11-40

After:

Screenshot_2025-11-24_14-14-07

This is what GPU-Viewer shows:

Screenshot_2025-11-24_14-19-46

Screenshot_2025-11-24_14-19-56

This is the output of `lspci -k | grep -EA3 'VGA|3D|Display'`:

```
system76@pop-os:~$ lspci -k | grep -EA3 'VGA|3D|Display'
00:02.0 VGA compatible controller: Intel Corporation Alder Lake-P GT2 [Iris Xe Graphics] (rev 0c)
	DeviceName: VGA compatible controller
	Subsystem: CLEVO/KAPOK Computer Alder Lake-P GT2 [Iris Xe Graphics]
	Kernel driver in use: i915
	Kernel modules: i915, xe
```

I'm concerned that we're losing the friendly name `Iris Xe Graphics`. I understand the bit in brackets is redundant in some cases (e.g. where it was just `[Intel Graphics]` on the gaze20 that Garrett showed earlier), but in this case, it's additional information that could be helpful to people. Is it possible to show that part?

FreddyFunk:
@jacobgkau improved the Intel quirk logic to address your latest feedback. also rebased this to the latest master

mmstick (dismissed):

cosmic-settings/src/pages/system/mod.rs
```diff
@@ -4,6 +4,8 @@
 #[cfg(feature = "page-about")]
 pub mod about;
 
+#[cfg(feature = "page-about")]
```
jacobgkau: I resolved the conflict with https://github.com/pop-os/cosmic-settings/pull/1597, but the spacing seems weird here. Do we want blank lines between each of these `#[cfg(feature = "page-about")]` blocks, or not? It seems like rustfmt isn't doing anything about it.

jacobgkau:
I pushed a commit with two small tweaks for Intel devices.

- One of them is to get rid of a non-existent/redundant second Intel device showing up on some systems, like the addw5 (this fixes a bug present in master). 
- The other further shortens the string by stripping out ` [Intel Graphics]` in case that's all that's in the bracket portion; that way, we can keep it when it contains names like Iris Xe, but not have the redundancy when it's just saying Intel again.

Current master:

Screenshot_2025-12-02_13-45-30

Before my last commit:

Screenshot_2025-12-02_13-42-09

After my last commit:

Screenshot_2025-12-02_16-18-13

jacobgkau:
I've added additional code to display the AMD GPU codename after the regular name, so we don't lose the friendly codename that was displayed before. I also removed the AMD GPU name from the CPU name, so it's not redundantly listed on the CPU line and its own line.

Before the last two commits:

Screenshot_2025-12-09_15-35-21

After:

Screenshot_2025-12-09_15-30-53

jacobgkau:
I'll resolve the conflict to rebase this after https://github.com/pop-os/cosmic-settings/pull/1623 is merged.

PR #1569, opened 2025-11-12
remove fractional refresh rates from the drop-down menu

FreddyFunk:
remove fractional refresh rates from the drop-down menu when a non-fractional refresh rate within 1Hz distance is available

before:
image

after:

image

### DIFF
--- a/cosmic-settings/src/pages/display/mod.rs
+++ b/cosmic-settings/src/pages/display/mod.rs
@@ -199,6 +199,8 @@ struct ViewCache {
     vrr_selected: Option<usize>,
     resolution_selected: Option<usize>,
     scale_selected: Option<usize>,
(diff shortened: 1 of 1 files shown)


mmstick:
I wonder if these are useful to have for applications that are sensitive to tearing if the refresh rate isn't slightly off for this particular display.

Drakulix:
> I wonder if these are useful to have for applications that are sensitive to tearing if the refresh rate isn't slightly off for this particular display.

Yeah, we shouldn't remove those. These are specifically different modes and a particular display might perform better with one, while having to interpolate with the other.

Possibly we could read out more timing information from the EDID and hide sub-optimal modes in the future, but the even refresh rates aren't guaranteed to be the correct ones.

Issue #1566, opened 2025-11-12
[Feature Request] Edit command and title for custom keybind after creation


Unkonwn-Crash:
Currently when editing a shortcut you are unable to modify the title or command it triggers. Inconvenient as I'm simply trying to edit(i.e capitalize first letter) and don't need to recreate the whole shortcut.
<img width="1306" height="492" alt="Image" src="https://github.com/user-attachments/assets/86566938-1ab2-45a1-bacc-1144e85654b2" />
I'd love to see the panel used for shortcut creation be used for shortcut editing. Like here:
<img width="1269" height="501" alt="Image" src="https://github.com/user-attachments/assets/a3278c1e-5c7a-4ec9-8c99-291cca4d348d" />
Side note, it makes more sense to be able to delete extra keybindings in the edit menu when creating than needing to click on it after the fact and delete any extra keybindings separately from the creation screen.

PR #1575, opened 2025-11-14
feat: modify custom binding action and description after creation

GenericConfluent:
Implement a suggestion from #1566

## Changes
- Use the `add_keybinding_context` for both creating by auto-filling from existing shortcut.
- Allow deleting binding from `add_keybinding_context` (Removed padding around binding input rows to make it look better with extra delete button, could use some adjusting)

https://github.com/user-attachments/assets/7c4bf5a5-c56a-47cd-b595-86e5467602fc


KodeBarista:
Can conflicts be resolved to merge this fix?

PR #1581, opened 2025-11-20
Some additional capslock options

ysblokje:
Dear devs, 

I use my capslock in a way that is not currently configurable so I created some additions.

so far I just added 2 new options: 

 - caps:escape_shifted_capslock
 - caps:none

### DIFF
--- a/cosmic-settings/src/pages/input/keyboard/mod.rs
+++ b/cosmic-settings/src/pages/input/keyboard/mod.rs
@@ -60,7 +60,7 @@ static CAPS_LOCK_OPTIONS: &[(&str, &str)] = &[
         "caps:ctrl_shifted_capslock",
     ),
     (
-        "Escape, Shift+Caps lock is regular Caps Lock",
+        "Escape, Shift+Caps Lock is regular Caps Lock",
         "caps:escape_shifted_capslock",
     ),
 ];


leviport:
That looks like 3 options

ysblokje:
> That looks like 3 options

You are quite right. I shall update the commit message.

leviport:
Thanks, and sorry for being pedantic :sweat_smile:

ysblokje:
> Thanks, and sorry for being pedantic 😅

I have no issues with that ;) 

So yes there are 3 options and the commit message reflects that now.

ysblokje:
Since it hasn't been reviewed I have added a few more options to my PR. I hope nobody minds. 
 I did leave out some of the options. I wasn't sure of the Azerty ones would make sense without actually selecting an Azerty keyboard layout.

maria-komarova:
Could I see the screenshot of the UI with updates?

leviport:
I just pushed a testing branch so CI would build it. It'll be on `popdev:testing-cosmic-settings-pr1581` once it's done building. I will post screenshots if I get to it before you have a chance to peek at it.

maria-komarova:
Thanks.

leviport:
Here's a screenshot with the new options:

Screenshot_2025-12-02_14-27-18

And if I squeeze the window down skinny, here's how those longer strings wrap:
Screenshot_2025-12-02_14-29-06

maria-komarova:
Great, thanks! For some reason the branch just wouldn't update for me so the screenshots help. 

I think it is too many options and the experience is overwhelming. We should try to reduce the number of options if we can. I'd also like to better understand how the following options behave exactly, what happens when you press the first two key combinations and what is the difference between the last three?

- Ctrl + Shift + Caps Lock
- Esc + Shift + Caps Lock
- Toggle Shift Lock (affect all keys)
- Shift with locking: Shift "pauses" Caps Lock
- Caps Lock acts as Shift with locking

leviport:
1. I don't think "Control, Shift+Caps Lock is regular Caps Lock" is working correctly. My understanding is that option should make Caps function as Ctrl as long as it's not preceded by Shift. It is merely acting as a regular Caps Lock currently.

2. I would expect "Escape, Shift+Caps lock is regular Caps Lock" to be pretty much the same, but acting as Esc instead of Ctrl. This one does seem to be working the way I'd expect. Shift + Caps activates/deactivates Caps Lock, while Caps alone acts as an Esc key. One issue with this mode is that the first "lock" is lowercase instead of capitalized to match the line above it.

3. I'm not quite making sense of how "Toggle Shift Lock" is supposed to be working. With that mode active, pressing Caps activates Caps Lock, which is deactivated either when I press Caps again, or do a single press on a Shift key. If I hold Shift and type a letter, an uppercase letter is printed and Caps Lock remains enabled. I struggle to imagine a scenario where this functionality would be desirable to anyone. I thought it would be more like a sticky keys mode, but this seems quite different.

4. "Shift with locking; Shift 'pauses' Caps Lock" seems indistinguishable from regular Caps Lock. Pressing Caps activates Caps Lock, and all letters are uppercase. Holding shift and pressing letters types lowercase letters. Is this just not working correctly, or is it supposed to do the same thing as the regular Caps Lock mode?

5. I'm also not understanding what the last one, "Caps Lock acts as Shift with locking" is supposed to do. It appears to just be setting Caps Lock, but in a weird state wherein holding Shift and typing letters does not type them in lowercase. I'm not sure what value that would have. Is it also not working correctly?

If we can't do anything about the weird functionality of these extra modes, I think I'd be satisfied slimming the new options down to "Disable" and "Escape, Shift+Caps Lock is regular Caps Lock" (but fix the capitalization on the latter) for this PR and leaving off the rest.

ysblokje:
Maybe my attempt to make the PR a bit more meaningful was not the best choice in hindsight. 

I checked my changes to be sure they at least do what they say.
(I assumed the manpage for xkeyboard-config would be correct but still). 

For reference: my keyboard is a mountain 60 with qwerty layout and setup as us-intl with deadkeys.
 1. works for me, might this be a layout issue? I could see myself using that one the future.   
 
 I will remove the bottom 3 and fix the capitalization.

mmstick (approved):

leviport:
I was worried that my work machine that I've been testing on might have been interfering, since I run some fairly heavily customized keyboard setups, so I switched to one of our Gazelle laptops with an unmodified en-US QWERTY layout. However I'm still seeing issues with some of the binding options:

- The Return key option just acts like a normal Caps Lock key. After consulting `man xkeyboard-config`, I'm not seeing a `caps:return` option in the Caps Lock behavior table:
Screenshot_2025-12-02_18-01-38

- Control shifted caps lock also isn't working on this Gazelle, and I'm also not seeing `caps:ctrl_shifted_capslock` in that table, so that could be the reason.

Other than those two, this seems to be working great!

maria-komarova:
Would it be possible to fix the order and copy for some of the options while we are on it? Caps Lock options never got a design review before so there are some inconsistencies with how options are presented in Alternate character key and Compose key.

The order and copy for the options:
- None
- Caps Lock [selected by default]
- Esc
- Backspace
- Ctrl
- Super
- Menu key
- Enter [pending QA approval]
- Num Lock
- Swap Esc and Caps Lock keys
- Super + Shift + Ctrl + Alt
- Ctrl, or Caps Lock when pressing Shift + Caps Lock [pending QA approval]
- Esc, or Caps Lock when pressing Shift + Caps Lock

ysblokje:
> I was worried that my work machine that I've been testing on might have been interfering, since I run some fairly heavily customized keyboard setups, so I switched to one of our Gazelle laptops with an unmodified en-US QWERTY layout. However I'm still seeing issues with some of the binding options:
> 
>     * The Return key option just acts like a normal Caps Lock key. After consulting `man xkeyboard-config`, I'm not seeing a `caps:return` option in the Caps Lock behavior table:
> 
> 
> Screenshot_2025-12-02_18-01-38
> 
>     * Control shifted caps lock also isn't working on this Gazelle, and I'm also not seeing `caps:ctrl_shifted_capslock` in that table, so that could be the reason.
> 
> 
> Other than those two, this seems to be working great!

- It seems the option for return was added in 2.43. 
- I have version 2.46 installed (Arch).
- Ubuntu 24.04, assuming Pop-os is still build on top of 24.04 has version 2.41

For reference: https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/blob/master/ChangeLog.md

Maybe some checks need to be done before presenting options? Or is that trying to be too flexible?

leviport:
Ah, that explains it. Yeah if there could be some checks before options are presented, that sounds like a pretty elegant solution to me. I'm not sure what's involved to implement that though. If it's a huge hassle, just commenting those two out with a `TODO` explaining the situation would probably be fine for now. I'm sure a new enough version will be in 26.04, which isn't terribly far away at this point.

ysblokje:
Over the past weeks have been thinking about the problem of having different versions of the `xkeyboard-config` and shouldn't there be a cleaner way of dealing with this. 

I am by no means an expert on this subject but would it be possible to just query that library? It has all the different options and a lot of translations already in place and from a quick glance it seems that the keyboard layouts are handled like that?

I do understand that the text is somewhat technical and cryptic (see previous comments) but it would be more agnostic to the underlying OS and/or libraries.

mmstick:
We should still support 24.04 after 26.04 is released. If checking the version at runtime is too difficult, an optional non-default cargo feature could be added to conditionally disable newer unsupported features on 24.04. We would simply set the feature conditionally in the build script.

ysblokje:
I have been trying to work out a different strategy last month that includes both the hard-coded options and the additional ones from the xkeyboard-config. With that in place it doesn't matter what the base OS has available it won't show what is not possible. 
I will try to fix it up to a presentable state and provide a screenshot of how that might look.

PR #1584, opened 2025-11-23
draft: storage page

FreddyFunk:
A lot of non-technical users default to look for these kind of information in the settings app like they are used to on macOS, Windows, iOS, Android, etc. in the settings app. 

Current preview:

https://github.com/user-attachments/assets/50c191a0-12bb-4b1a-be22-8f274386f91b

Looking for critical feedback to get this to a state where this might be acceptable.

Todo list to get this out of the "draft" state:
- [x] fast, asynchronous and parallelized file system scans
- [x] flatpak app overview
- [x] flatpak app details view
- [x] file system scan cache
- [ ] preserve file system scan cache during cosmic-settings restarts
- [ ] improve "Storage Overview" by getting showing used & available space immediately and then filling the used space with the storage categories
- [ ] support for multiple storage media
- [ ] support for multiple partitions or shrinked partitions that do not fill the entire storage media space
- [ ] support for multiple users with multiple home directories on one system
- [ ] define categories
- [ ] define sub-categories for "System"
- [ ] define sub-categories for "Home"
- [ ] define sub-categories for "Applications"
- [ ] define sub-categories for "Other"

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1713,6 +1713,7 @@ dependencies = [
  "indexmap 2.12.0",
  "itertools 0.14.0",
  "itoa",
+ "jwalk",
  "libcosmic",
  "locale1",
  "locales-rs",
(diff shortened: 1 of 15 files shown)


mmstick:
Needs UX approval from @maria-komarova with Figma mockups.

PR #1595, opened 2025-11-30
provide the option to apply RON theme files via the cli

FreddyFunk:
just run 'cosmic-settings --import-theme /path/to/theme.ron' to apply a theme from the cli without launching the COSMIC settings app GUI. This is especially helpful for anyone using dotfiles.

### DIFF
--- a/cosmic-settings/src/pages/desktop/appearance/commands.rs
+++ b/cosmic-settings/src/pages/desktop/appearance/commands.rs
@@ -3,9 +3,10 @@
 
 //! CLI commands for theme import and export.
 
-use cosmic::cosmic_config::CosmicConfigEntry;
-use cosmic::cosmic_theme::{Theme, ThemeBuilder, ThemeMode};
+use cosmic::cosmic_theme::ThemeBuilder;
 use std::path::Path;
+
+use super::theme_manager::Manager;
 
 /// Import a theme from a RON file and apply it.
 ///
@@ -38,41 +39,24 @@ pub fn import_theme(path: &Path) -> color_eyre::Result<()> {
     let is_dark = builder.palette.is_dark();
     let mode_str = if is_dark { "dark" } else { "light" };
 
-    if let Ok(mode_config) = ThemeMode::config() {
-        let mut mode = ThemeMode::get_entry(&mode_config).unwrap_or_default();
-        if mode.is_dark != is_dark {
-            let _ = mode.set_is_dark(&mode_config, is_dark);
+    let mut manager = Manager::default();
+
+    if manager.mode().is_dark != is_dark {
+        if let Err(err) = manager.dark_mode(is_dark) {
+            return Err(color_eyre::eyre::eyre!(
+                "Failed to set {} mode: {:?}",
+                mode_str,
+                err
+            ));
         }
     }
 
-    let (theme_config, builder_config) = if is_dark {
-        (Theme::dark_config(), ThemeBuilder::dark_config())
-    } else {
-        (Theme::light_config(), ThemeBuilder::light_config())
-    };
-
-    if let Ok(config) = builder_config {
-        builder.write_entry(&config).map_err(|e| {
-            color_eyre::eyre::eyre!("Failed to write theme builder config: {:?}", e)
-        })?;
-    } else {
-        return Err(color_eyre::eyre::eyre!(
-            "Failed to get {} theme builder config",
-            mode_str
-        ));
-    }
-
-    let theme = builder.build();
-    if let Ok(config) = theme_config {
-        theme
-            .write_entry(&config)
-            .map_err(|e| color_eyre::eyre::eyre!("Failed to write theme config: {:?}", e))?;
-    } else {
-        return Err(color_eyre::eyre::eyre!(
-            "Failed to get {} theme config",
-            mode_str
-        ));
-    }
+    manager
+        .selected_customizer_mut()
+        .set_builder(builder.clone())
+        .set_theme(builder.build())
+        .apply_builder()
+        .apply_theme();
 
     println!(
         "Successfully imported {} theme from: {}",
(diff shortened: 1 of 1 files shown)


mmstick:
Could this be a subcommand of the appearance page instead? A global arg might be messy if we start adding more of these over time. Perhaps `cosmic-settings appearance import THEME_CONFIG_PATH`. And `cosmic-settings appearance export THEME_CONFIG_PATH`.  See `cosmic-settings -h`.

FreddyFunk:
> Could this be a subcommand of the appearance page instead? A global arg might be messy if we start adding more of these over time. Perhaps `cosmic-settings appearance import THEME_CONFIG_PATH`. And `cosmic-settings appearance export THEME_CONFIG_PATH`. See `cosmic-settings -h`.

this sound like great idea! 

 new CLI usage

```bash
  # Import a theme
  cosmic-settings appearance import THEME_CONFIG_PATH

  # Export the current theme
  cosmic-settings appearance export THEME_CONFIG_PATH

  # Still works - opens the appearance page in GUI
  cosmic-settings appearance
```

cosmic-settings/src/main.rs
```diff
@@ -191,6 +217,127 @@ pub fn main() -> color_eyre::Result<()> {
...
+
+    println!(
+        "Successfully exported {} theme to: {}",
+        mode_str,
+        path.display()
+    );
+
+    Ok(())
+}
+
```
mmstick: Can these two functions be moved to the the appearance module somewhere? Maybe in its own commands module.

FreddyFunk:
addressed the feedback and also updated the logic so that the CLI commands use theme_manager::Manager for consistent behavior with the UI

mmstick (approved):

ada4a:
(a complete rando here)

Imo this feels out of place even as a subcommand: the logic seems pretty independent from that of `cosmic-settings`[^2]. Maybe this could be implemented as (a) separate command(s) instead, called something like `cosmic-apply-theme`/`cosmic-export-theme`, similarly to KDE Plasma's `plasma-apply-*`[^1] commands?

[^1]: https://discuss.kde.org/t/cli-commands-to-switch-between-theming-settings/15544/2
[^2]: it does make use of `theme_manager::Manager`, so the new command would, for now, need to be implemented as a new binary in `cosmic-settings`, but maybe this is a sign to extract `theme_manager` into a separate, reusable crate

FreddyFunk:
Not sure if I agree. A new binary feels a bit over engineered in my opinion. Keeping it as a settings sub-command keeps the theming logic closer together as a single point of truth. But if necessary, I'd be open to implement this as a dedicated binary. Is there any opinion on this from the maintainers?

ada4a:
Maybe.. the thing that bothers me is that the new subcommand goes inside `PageCommand`, even though it's not, well, a page command. One could imagine moving all the page commands under a subcommand like `cosmic-settings open (input, desktop, etc.)`, and adding new subcommands `cosmic-settings import/export`, but that might make the (probably) common case of opening a specific page a bit too verbose...

Yeah let's see what the maintainers think

FreddyFunk:
@mmstick could you please take a look at this again when you have time? :)

mmstick:
@FreddyFunk Needs approval from QA.

jacobgkau (approved):
The import & export subcommands are working as expected, as are the page commands for this and other pages. The `cosmic-settings appearance -h` output looks good.

Attempting to export to a bad path (e.g. one we don't have permission for, or one in a directory that doesn't exist) gives an error message that's appropriate enough (showing the underlying `Permission denied` or `No such file or directory` as part of the output). Attempting to import a nonexistent file does the same. Attempting to import a malformed RON file gives an appropriate message (`Unexpected end of RON`, `Expected float`, etc) message, which is good.

It looks like there are some validation things that aren't enforced when importing, like color floats being able to go above 1 when they appear to be using a 0-1 scale; but I can already do the same thing by clicking `Import` on the released version of COSMIC Settings, so that's not a regression with this feature.

<details>
<summary>Regression testing passed:</summary>

## Checklist

### Network & Wireless

#### Wi-Fi

- [X] Able to disconnect from and connect to a network.

### Desktop

#### Wallpaper

- [X] Clicking a different wallpaper switches to it immediately
- [X] Leaving the page and coming back does not change the wallpaper

#### Appearance

- [X] Open a COSMIC app and a GNOME app. When switching between Dark and Light mode, the COSMIC app switches immediately, and the GNOME app switches after restarting.
- [X] Changing the accent color works.
- [X] Changing the roundness style works and does not affect other aspects of the theme (e.g. color mode).
- [X] Changing the interface density works.

#### Panel

- [X] Enable tiling. "Automatically hide panel" takes effect immediately.
    - Not a regression.

#### Dock

- [X] Enable tiling. "Automatically hide dock" takes effect immediately.
- [ ] Disable tiling; maximize a window. "Autommatically hide dock" takes effect immediately.
    - Not a regression.
- [X] Changing position on screen works.
- [X] All Style settings take effect as expected.

### Displays

- [X] Plug in a second display. Displays can be rearranged via dragging.

### Sound

- [X] Page works as expected.

### Power & Battery

- [X] Changing power mode takes effect and is reflected by top panel applet.
    - Change in applet doesn't reflect in Settings until re-entering the page; not a regression.

### Input Devices

#### Touchpad

- [X] All click behavior settings work as expected.
- [X] All scrolling settings work as expected.

### Applications

#### Default Applications

- [X] Changing the default terminal affects Super-T shortcut.

- [X] Global Shortcuts in X11 Applications works (in e.g. Discord).

### Time & Language

#### Date & Time

- [X] Changing time zone works.
- [X] Toggling 24-hour time takes effect in Settings preview, on top panel, and on the lock screen.
- [X] Toggling seconds takes effect in Settings preview and on top panel.
- [X] Changing first day of week takes effect in Calendar applet.
- [X] Turning date on/off for applet works.

### System & Accounts

#### About

- [X] Changing hostname works (check terminal prompt).
</details>

PR #1604, opened 2025-12-02
chore(sound-subscription): disable libcosmic default features

mmstick:
The subscription crate is used as a library by applets, which currently does not use default features. As a library this doesn't need default features either since they'll get enabled (if needed) by the application that imports it.


mmstick:
Since this does not affect the lockfile (no features removed or added) there should be no change in behavior for cosmic-settings.

git-f0x:
This will conflict with #1567, but I'll remove that commit from there. `log` can also be removed as a dependency of the sound subscription, since it's unused.

mmstick:
It'd be a good idea to do another sweep with [cargo-machete](https://github.com/bnjbvr/cargo-machete) soon.

jackpot51 (dismissed):

wash2 (approved):

PR #1606, opened 2025-12-02
feat(shortcuts): add accessibilty shortcuts subpage w/ zoom & dezoom

### LINKED ISSUE #1602: [Feature Request] Configure/Change/Remove zoom keybinds
Hello, new COSMIC User here (from Sway). While configuring the keybinds, I noticed that there is no direct way to change zoom keys, which in my case are "Super+." and "Super+," . These used to be super central to my workflow and this is a mild annoyance.

So far I have looked in:
 - the input -> keybinds screen
 - the accessibility -> zoom screen
 - looking for config files with the pattern 'keys', 'zoom', and 'keyboard', and manually looked in `com.system76.CosmicComp/v1/accessibility_zoom`, `com.system76.CosmicSettings.Shortcuts/v1/custom` and `com.system76.CosmicSettings.Shortcuts/v1/system_actions`
 
I also looked for similar issues and have found nothing.
 
Please let me know if I have missed it 🙂.

If not, would it be possible/easy to implement ? (if not already planned)

I'd be willing to make a PR if needed, although I do not have much time.

Thanks!

payolin:
Closes #1602.

This PR
- Adds an "accessibility" subpage under input -> keyboard -> shortcuts
- Adds a bind entry for the Zoom action in said subpage
- Adds a bind entry for the Unzoom action in said subpage

It could be argued that these didn't need their own subpage. However, I felt it didn't fit in System, because it's only used to bind actions in the SystemAction enum. Other categories didn't feel adequate either, except maybe for navigation because I guess you could navigate around your screen by zooming/dezooming.

Please lmk if any change is to be made by me and thank you to reviewers and maintainers for their time and effort!

### DIFF
--- a/cosmic-settings/src/app.rs
+++ b/cosmic-settings/src/app.rs
@@ -384,12 +384,9 @@ impl cosmic::Application for SettingsApp {
                         return page.update(message).map(Into::into);
                     }
                 }
-                #[cfg(feature = "page-input")]
-                crate::pages::Message::AccessibilityShortcuts(message) => {
-                    if let Some(page) = self
-                        .pages
-                        .page_mut::<input::keyboard::shortcuts::accessibility::Page>()
-                    {
+                #[cfg(feature = "page-about")]
+                crate::pages::Message::About(message) => {
+                    if let Some(page) = self.pages.page_mut::<system::about::Page>() {
                         return page.update(message).map(Into::into);
                     }
                 }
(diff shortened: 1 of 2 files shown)


mmstick:
This adds an accessibility shortcuts page that has only two shortcuts: `ZoomIn` and `ZoomOut`

maria-komarova (approved):

mmstick (approved):

jacobgkau (approved):
There are two additional improvements that would be good to add:

1. Clarify the name of these new shortcuts. They're currently just `Zoom In` and `Zoom Out`, which, while maybe clear when browsing under the Accessibility page, could lead people to think they apply to the zoom function in other apps when they don't, especially if found via searching.
2. Make the hint text on the Accessibility settings page update to show the actual configured shortcuts, instead of being hard-coded to the default shortcuts (or at least clarify the text to indicate it's only talking about the defaults).

But this is working as-is, so if UX is good with it, then we can start with this.

payolin:
Nice, thank you !

I'll try to make a follow-up PR implementing the changes mentioned by @jacobgkau when I can. For now at least the option is here.

maria-komarova:
> Clarify the name of these new shortcuts. They're currently just Zoom In and Zoom Out, which, while maybe clear when browsing under the Accessibility page, could lead people to think they apply to the zoom function in other apps when they don't, especially if found via searching.

I think it is fine since it is easy to see what is happening once trying to use magnification shortcuts and we don't list per app shortcuts in Settings. But if we want to be crystal clear then we can use "Zoom in (magnifier)" and "Zoom out (magnifier)".

> Make the hint text on the Accessibility settings page update to show the actual configured shortcuts, instead of being hard-coded to the default shortcuts (or at least clarify the text to indicate it's only talking about the defaults).

Good point, we should make sure the hint updates.

PR #1616, opened 2025-12-06
Russian translation for .desktop file

FAlexei:


jackpot51 (approved):

ada4a:
"Hacтройки" is the more commonly used name for this, no?

FAlexei:
@ada4a, "Параметры" is already used across all components in cosmic.
As far as I know, Windows uses the same name as well.

ada4a:
> As far as I know, Windows uses the same name as well.

Not on Windows 11 at least...

IMG-20260106-WA0001

Android calls it "Настройки" as well.

>  "Параметры" is already used across all components in cosmic.

Imo that's just a mistake repeated multiple times, which doesn't make it less of a mistake... (This sounds confrontational however I phrase it – apologies for that, that was not my intention)

FAlexei:
@ada4a, it had always been called "Параметры" before they fired all the translators and started using Bing Translator, which just pushes the first available translation.
image

KDE Plasma uses "Параметры системы" right now.
image

Both variants are correct. There are no mistakes.

Issue #1573, opened 2025-11-13
[Bug] No default selection visible for Click Behavior & Scrolling in Touchpad settings


kodebarista:
**Issue/Description:**
In `Settings > Input Devices > Touchpad`, the UI does not show any default selection for `Click Behavior` and `Scrolling` options, even though a default action is already applied internally.

**Version:**
`Source          : cosmic-settings-1.0.0~beta.5-1.fc43.src.rpm`

**Steps to Reproduce:**
- Open Settings
- Navigate to `Input Devices > Touchpad`
- Observe `Click Behavior` and `Scrolling` options

**Expected Behavior:**
The default option for `Click Behavior` and `Scrolling` should be visibly selected in the UI to reflect the active configuration.

**Other Notes:**

<img width="689" height="463" alt="Image" src="https://github.com/user-attachments/assets/ab38cdda-ec19-45ff-8839-03c140c7e6ee" />

jacobgkau:
Confirmed this issue on fresh installations on the pang15 and darp11. This is what it looks like by default:

<img width="2560" height="1600" alt="Image" src="https://github.com/user-attachments/assets/6dd8ba92-03d1-4317-ba7c-552c7d86d2d1" />

PR #1620, opened 2025-12-08
fix(input): install defaults for input_touchpad config

wash2:
fixes https://github.com/pop-os/cosmic-settings/issues/1573. Or should cosmic-comp should install the defaults for its config instead?


wash2:
> Is `/usr/share/cosmic/com.system76.CosmicComp/` where it should be installed to?

Ya, that is where it would read the system default for the cosmic-comp touchpad input config from

jacobgkau (approved):
This does make the defaults show up instead of being blank.

Changing scroll mode works. Changing from multitouch to click zones does not work, but that's not a regression. Also, I'm not seeing a `Reset to defaults` button appear on this page after changing any settings, but again, that's not a regression.

PR #1623, opened 2025-12-09
chore: update dependencies

mmstick:


jackpot51 (dismissed):

cosmic-settings/src/pages/input/keyboard/shortcuts/mod.rs
```diff
@@ -732,6 +732,7 @@ fn localize_action(action: &Action) -> String {
             SystemAction::PlayPrev => fl!("system-shortcut", "play-prev"),
             SystemAction::PowerOff => fl!("system-shortcut", "poweroff"),
             SystemAction::Screenshot => fl!("system-shortcut", "screenshot"),
+            SystemAction::Suspend => fl!("system-shortcut", "suspend"),
```
jacobgkau: It looks like this is supposed to show up as a shortcut on the System page of the keyboard shortcut settings? I'm not seeing it in there. Is it supposed to be visible, or is this just used for something else?

mmstick: It makes it possible to localize, but indeed it needs to be manually added to that category to appear there.

mmstick: It makes it possible to localize, but indeed it needs to be manually added to that category to appear there.

wash2 (dismissed):

cosmic-settings/src/pages/input/keyboard/shortcuts/mod.rs
jacobgkau: 93164e6 makes it show up in search without having to be set first, like all other default shortcuts.

mmstick:
Nothing I can do here. Seems to be expected behavior? https://github.com/pop-os/cosmic-settings-daemon/pull/55#issuecomment-3628533033. Once cosmic-comp updates its dependency on https://github.com/pop-os/cosmic-settings-daemon/pull/55 it will be able to handle it.

jacobgkau:
I tried bumping cosmic-settings-daemon in cosmic-comp (`bump-settings-daemon` branch), but it's still not working.

mmstick:
It would need to be the `cosmic-settings-daemon-config` crate specifically

PR #1626, opened 2025-12-11
chore: format code

ada4a:
fixes CI

PR #1627, opened 2025-12-11
ci: set apt to non-interactive

ada4a:
Currently, apt tries to initialize various frontends, but eventually realizes that none of them work, as it's run on CI. This commit stops it from trying in vain.


ada4a:
The rustfmt error is spurious; it'll be fixed by https://github.com/pop-os/cosmic-settings/pull/1626

PR #1653, opened 2025-12-16
feat: add control for alerts volume/mute (fixes #1017)

### LINKED ISSUE #1017: Add toggle to disable system sounds
I and many others don't like random functions on our system making sound, such as the volume slider. It would be nice if there was a toggle to disable all of the sounds.

cgevans:
Using an approach based on the System Sounds setting in pavucontrol, this gets and sets route-settings metadata that adjusts the volume of media with the Notification media-role.

This works for me.  However, I suspect that it may only be working because pavucontrol has already added the right metadata for this code to modify.   I'm also not sure that my handling of finding the metadata is the best approach.  I'm leaving this as a draft for now until I can test it in a way that avoids the potential for the route-settings already being there.

Some other points I'm unsure of:

- Since there was already a commented out "Alerts" section, I used that.  I'm not sure if there's a preferred term here for Cosmic (if I recall, Pipewire uses the Alert media-role for something else).
- I added the slider at the top here, but it could equally well go at the bottom.  The awkwardness of it being at the bottom is that it would be away from the other volume sliders, and underneath "Device profiles"; as Device profiles currently has no heading, and applies to both Output and Input, the arrangement would be a bit awkward, as it might make Device profiles seem like it was part of Input.  There may be better options here I'm not thinking of at the moment.


cgevans:
I've now tested this with a new user, and it seems like it works without a problem.  It will create the metadata if it doesn't already exist, with a 100% volume setting (the current default anyway).  My only concern is if there's no route-settings metadata at all, but I'm not sure if that's possible?

cgevans:
From my perspective: I hadn't thought about having the control in the outputs section; that could make sense, especially as there is currently only one output volume in output.   But there are perhaps wider questions:

- Pipewire has multiple media roles that could have volumes set.  Notification is important because Cosmic itself uses them, but some systems also use things like Alarm or Music.  I don't think any of them are necessarily important here, however.
- Per-app volume adjustment is of course possible, and things like pavucontrol support it.
- Right now the sounds panel has only one output and one input controllable, but it could alternatively make sense to have controls for all/some outputs and inputs, with the user might be using.  If there were multiple volumes, having the alerts there might be confusing.  But while I often do actually need to control multiple volumes, I feel like this might be more something for separate, specialized applications, with the settings here left for default output/input; that might also apply to per-app adjustment.
- The Alerts Volume here is an adjustment to the output volume, so alerts become quieter when the output volume is turned down.  This might actually suggest that having it in the Output section would be preferable.  It also might be best to make this clear somehow, as it differs from, eg, the Android role-based volumes.  I'm just not sure of what wording would be best.  "Alert volume adjustment" doesn't quite work for me.

jacobgkau:
> Right now the sounds panel has only one output and one input controllable, but it could alternatively make sense to have controls for all/some outputs and inputs, with the user might be using. If there were multiple volumes, having the alerts there might be confusing. But while I often do actually need to control multiple volumes, I feel like this might be more something for separate, specialized applications, with the settings here left for default output/input; that might also apply to per-app adjustment.

Yeah, I think the decision's been made to generally keep it to one default output device and one default input device for now. It's possible to configure volumes for various devices, but COSMIC Settings currently only aims to configure volumes for the devices set as the defaults. (Some apps can use and configure non-default devices themselves.)

> The Alerts Volume here is an adjustment to the output volume, so alerts become quieter when the output volume is turned down. This might actually suggest that having it in the Output section would be preferable. It also might be best to make this clear somehow, as it differs from, eg, the Android role-based volumes. I'm just not sure of what wording would be best. "Alert volume adjustment" doesn't quite work for me.

This is a good thing to keep in mind, but I don't think we need to overthink it, personally. Somehow visually nesting the `Alerts volume` slider underneath the `Output volume` slider would make what you're talking about clear. But I don't think adding the word "adjustment" would make clear what you're trying to communicate (someone reading it with no context could just think you mean adjusting the alert volume in the general sense, not having anything to do with the output volume).

cgevans:
Here it is directly under the output volume:

image

And here it is with a \t before the title to indent it:

image

I think the indentation works, but I'm not sure it's used anywhere else, so it might be something to be cautious about.

cgevans:
Also, I ran cargo fmt, but it caused unrelated changes to `cosmic-settings/src/pages/system/info.rs`, the same formatting that caused the CI failure for the previous commit.

jacobgkau:
Thanks for posting screenshots of what those other configurations would look like. I'll wait for @pop-os/ux to check in on this.

maria-komarova:
As to how we can call it, maybe "System alerts" or "System sounds"?

Regarding placement we probably shouldn't overthink it. I checked what other systems are doing in their UI and system alerts are in a separate section even though they are controlled by the output volume slider. So shouldn't be an unexpected behavior. That said I don't have many objections to placing the slider under Output other than the examples of other systems and the potential future features. For example, options for alert sounds. But since there is no definite plan to have those right now it's probably okay to keep the setting under Output. No need to indent though since this raises more questions than serves to explain.

PR #1669, opened 2025-12-18
i18n(fr): add translations to desktop entries

AnonAlly:


AnonAlly:
Should be ready for review ^^.

PR #1727, opened 2026-01-02
chore: update cosmic-randr dependencies

craftyguy:
Fixes #1716

PR #1728, opened 2026-01-03
Update branches crate to fix build on pp64le

ligenix:
The branches crate previous to 0.4.4 breaks builds on ppc64le arch.


jacobgkau:
https://github.com/pop-os/cosmic-settings/pull/1727 updates all dependencies and may obsolete this PR.

PR #1738, opened 2026-01-07
improv(network): focus the secure input

### LINKED ISSUE #1737: Wi-Fi password dialog doesn't focus itself
I noticed this while testing https://github.com/pop-os/cosmic-settings/pull/1237.

When connecting to a secured WiFi network for the first time, the password dialog doesn't get focused when it appears. This means seeing the dialog pop up and starting to type the password doesn't work. (Prior to https://github.com/pop-os/cosmic-settings/pull/1237, typing did nothing; after https://github.com/pop-os/cosmic-settings/pull/1237, it will search Settings.) The user has to manually focus the password dialog before it will work, either by clicking inside of it or pressing Tab until it's focused.

Ideally, the password input box should be focused automatically when the dialog appears.

wash2:
fixes #1737

### DIFF
--- a/cosmic-settings/src/pages/networking/vpn/mod.rs
+++ b/cosmic-settings/src/pages/networking/vpn/mod.rs
@@ -785,6 +785,7 @@ impl Page {
             }
             Message::VpnDialogError(vpn_dialog) => {
                 self.dialog = Some(vpn_dialog);
+                return task::message(Message::FocusSecureInput);
             }
             Message::SecretAgent(e) => match e {
                 nm_secret_agent::Event::RequestSecret {


jacobgkau (commented):
This gets the WiFi password dialog focused automatically when it opens up. This actually blocks interacting with the search box or anything else in the background window while the WiFi password dialog's open, which seems fine. Interaction resumes as normal once the dialog's submitted or cancelled.

I see the PR's possibly trying to do the same thing for the VPN page. A test VPN file that I'd imported a while back (that doesn't actually lead to a valid VPN server) shows a "Failed to connect to VPN" dialog box that includes a password field; that field is still unfocused by default, and typing here still results in searching the background window instead. @wash2 Should this dialog also get the same treatment?

Image

wash2:
> This gets the WiFi password dialog focused automatically when it opens up. This actually blocks interacting with the search box or anything else in the background window while the WiFi password dialog's open, which seems fine. Interaction resumes as normal once the dialog's submitted or cancelled.
> 
> I see the PR's possibly trying to do the same thing for the VPN page. A test VPN file that I'd imported a while back (that doesn't actually lead to a valid VPN server) shows a "Failed to connect to VPN" dialog box that includes a password field; that field is still unfocused by default, and typing here still results in searching the background window instead. @wash2 Should this dialog also get the same treatment?
> Image

Ya, I think the failed to connect dialog was overlooked on my part. I'll push an extra change that focuses this dialog as well.

mmstick (approved):

PR #1747, opened 2026-01-09
fix(a11y): listen to screen reader changes from org.a11y.Bus

mmstick:
Enables the screen reader toggle to toggle when using the shortcut.


jacobgkau (approved):
This gets the `Screen reader` toggle on the Accessibility page of Settings toggling in real-time when the shortcut's used to turn it on/off. The shortcut still works as expected when Settings isn't open. The rest of the Accessibility page still works as expected.

PR #1759, opened 2026-01-15
fix: fallback to saving pw in system if secret agent channel is closed

wash2:


leviport:
Cheesing with a double QA review so I can start building ISOs sooner

PR #1779, opened 2026-01-23
Update com.system76.CosmicSettings.Users.policy ITA

lucamosca1:


lucamosca1:
Fixed (apopogies)

lucamosca1:
Added also com.system76.CosmicSettings.StartupApps.desktop

Issue #1692, opened 2025-12-22
Wallpaper slideshow resets the duration


abimagnus:
**Describe the bug**
If i choose the option for wallpaper slideshow and change the image every 30 minutes its resets backs to 5 minutes

**To Reproduce**
1. Right click desktop > change  wallpaper> enable slideshow > change image every> 30 minutes>close it
2. Again Right click desktop > change  wallpaper> the change  image every resets to 5 minutes

**Expected behavior**
The change image every option duration should stick

**Screenshots**

<img width="940" height="251" alt="Image" src="https://github.com/user-attachments/assets/a2a94e26-bcb8-4fb1-b18a-675f0a6a2ebf" />

**Desktop :**
 - OS: Pop!_OS 24.04]

PR #1780, opened 2026-01-23
Fix rotation frequency of wallpaper slideshow resetting

SammyMcFly:
The additions/changes proposed fix the behavior that the "rotation_frequency" variable is not stored. As a result the rotation frequency of the wallpaper slideshow reset every time the settings window was closed.

This PR fixes the bug reported in #1692

### DIFF
--- a/cosmic-settings/src/pages/desktop/wallpaper/config.rs
+++ b/cosmic-settings/src/pages/desktop/wallpaper/config.rs
@@ -267,7 +267,10 @@ impl Config {
     /// # Errors
     ///
     /// Returns an error if the on-disk configuration could not be updated.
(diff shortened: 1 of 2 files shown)


SammyMcFly:
I just applied the suggested format changes.

SammyMcFly:
> I did notice the slideshow starts over when entering the page

Do you mean the timer?

jacobgkau:
> Do you mean the timer?

No, I mean the slideshow. It goes back to the first image every time I enter the page. (Again, not a regression, already happening before this PR.) I'd expect it to just continue the slideshow, resetting the timer (but not necessarily changing the image immediately) when the time setting's changed.

SammyMcFly:
Interestingly this does not happen on my PC. For me the image is kept when I enter the settings page or change wallpaper settings.

PR #1785, opened 2026-01-24
fix: increased hard-coded wallpapers limit from 100 to 500

### LINKED ISSUE #1356: Feature Request: Allow the Wallpapers section to use all of the available images inside a folder
When I go to the Wallpapers page and I click "open new folder", cosmic-settings doesn't use all of the selected folder's images, instead it only grabs a few. 
This is problematic if you have a lot of wallpapers and are using the slideshow feature, as you will quickly cycle through all of your wallpapers. 
Removing this limit or making it a toggle (in case adding the wallpapers is too performance intensive) would be a small but massive improvement. 

I am using cosmic-settings version: 1.0.0.alpha.7.r140.gbd7fc2a-1 from the Chaotic AUR.

thegabriele97:
### Increase wallpaper directory load limit

This PR raises the hard-coded limit for loading wallpapers from a directory from **100 to 500**.

fixes: #1356

The current limit of 100 is unnecessarily restrictive for users who use folders as wallpaper sources (e.g. slideshow-style setups) and is not configurable or user-visible, making the behavior appear unintended.

This change does not increase peak resource usage:
- loading is already throttled via `buffered(4)`
- images are processed incrementally

The increase improves usability while keeping the existing design and safeguards intact.

### DIFF
--- a/pages/wallpapers/src/lib.rs
+++ b/pages/wallpapers/src/lib.rs
@@ -134,7 +134,7 @@ pub async fn load_each_from_path(
         })
         .buffered(4)
         .filter_map(|value| async { value.ok().flatten() })
-        .take(100);
+        .take(500);
 
     Box::pin(future)
 }


mmstick:
We should hold off on this until libcosmic has been rebased on iced 0.14, and cosmic-settings has also been rebased on that rebase of libcosmic. 500 images uses a lot of VRAM and CPU in the current version of iced used.

KodeBarista:
Can this PR be reviewed now?

mmstick (approved):

jacobgkau (approved):
It appears from my testing that https://github.com/pop-os/cosmic-settings/issues/1356 didn't affect slideshows; cosmic-bg doesn't rely on this particular COSMIC Settings code when it loads a directory's contents for a slideshow. But it did affect display of the images in COSMIC Settings (which is needed for picking an individual image instead of using a slideshow).

This change increases the cap for wallpapers displayed in Settings from 100 to 500 (and the cap does still apply at 500). I don't see any performance degradation with 500 previews on the page.

All settings on the Wallpaper page still work as expected. The only code change is on this page, so full regression testing of the rest of the pages isn't necessary.

PR #1790, opened 2026-01-28
Avoid creating dropdowns for empty default app categories

istekhar8966:
ONLY ON ARCH BASED SYSTEM's

When there is no application installed for a default app category (like Music or Video), the settings app still tries to show a dropdown.
Clicking on it causes the settings app to crash on Wayland.
This change prevents creating the dropdown when the app list is empty and shows a simple text instead.


istekhar8966:
Thanks for the review!
I’ll update the label to sentence case and switch it to Fluent string.

jacobgkau (approved):
Looks good now:

Image

Regression testing for the default apps page passed (setting other, installed default apps still works).

Thank you!

PR #1796, opened 2026-01-29
Fix features

jackpot51:
This ensures cosmic-settings is as portable as possible


wash2 (approved):

jackpot51:
I made two more changes to ensure this is building for Redox OS, at least. It will likely also build on Windows/macOS for use of the appearance page

PR #1813, opened 2026-02-02
Style focus delay field as a regular text input

istekhar8966:
The “Focus follows cursor delay in ms” field looks like a label by
default. Apply the default text input style so it clearly appears and
behaves like an input field.

before:

before

After:

after


mmstick:
It would not be proper to override the style of the editable input widget. The style that it uses is already the correct style from our design docs. This widget was constructed with `cosmic::widget::editable_input()` and by design it [constructs a TextInput widget with the EditableText style](https://github.com/pop-os/libcosmic/blob/3e78eb238159d90956e85e95e868164671b649f6/src/widget/text_input/input.rs#L76). Using a different widget would require design approval first, and then using the correct widget function for that variant instead of overriding the style.

PR #1839, opened 2026-02-12
fix(sound): only consider active routes when setting volume and mute-ness

mmstick:
This should fix the audio device changing when the volume is modified. Will require syncing the same fix in cosmic-applets to get the fix there as well.


mmstick:
@jacobgkau An updated `pw-dump` from that one before and after audio dysfunctions might be helpful.

PR #1860, opened 2026-02-19
feat: set default archive app in default applications page

thegabriele97:
PR supporting https://github.com/pop-os/cosmic-files/pull/1555

----------------------------------


wash2 (approved):

jacobgkau (changes requested):
The sole remaining problem with this is that, on upgrade, Settings shows COSMIC Files as the default application, but both Firefox and COSMIC Files actually treat File Roller as the default. If I change the default application to File Roller and then back to COSMIC Files, then they respect COSMIC Files as the default. The desync needs to be fixed and the default needs to actually be COSMIC Files on upgrade.

thegabriele97:
> the default needs to actually be COSMIC Files on upgrade.

Mmmh how can this be achieved? I mean, "forcing" cosmic files to be the default in all the supported archive mime types on upgrade

thegabriele97:
I have a question about how Cosmic Settings handles multiple MIME types.

If I set "Cosmic Video Player" as the default under the Video category, and then manually change a specific MIME type (for example, video/x-matroska) to use mpv, Cosmic Video Player still appears as the default app for Videos in the settings.

It seems like the same issue occurs with the new Archive category: Settings shows Cosmic Files as the default application, even when a specific MIME type is actually being opened with Firefox.

PR #1878, opened 2026-02-26
Notifications settings page

### LINKED ISSUE #97: cosmic ignores --expire-time when passed to notify-send
If I pass --expire-time=60000 or any other value to notify-send, the notification disappears after a few seconds.

joshuamegnauth54:
Closes: pop-os/cosmic-notifications#97 pop-os/cosmic-notifications#83

I implemented a settings page for COSMIC Notifications so that it is easier to edit its config. COSMIC Notification already exposes a config to tweak useful options like the maximum time a notification may stay on the screen. Currently, there isn't a nice way to edit this config which can be confusing to end users.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1583,6 +1583,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "cosmic-notifications-config"
+version = "0.1.0"
+source = "git+https://github.com/pop-os/cosmic-notifications#b58fb29bd7e5898e4252ecb45d5bebfdecdf7cca"
+dependencies = [
+ "cosmic-config",
+ "serde",
+]
+
+[[package]]
 name = "cosmic-panel-config"
 version = "0.1.0"
 source = "git+https://github.com/pop-os/cosmic-panel#6a783a18c3079955ff4cde1d00eaadf8e77b869c"
@@ -1678,6 +1687,7 @@ dependencies = [
  "cosmic-dbus-networkmanager",
  "cosmic-idle-config",
  "cosmic-mime-apps",
+ "cosmic-notifications-config",
  "cosmic-panel-config",
  "cosmic-protocols 0.2.0",
  "cosmic-randr",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,9 @@ git = "https://github.com/pop-os/cosmic-idle"
 [workspace.dependencies.cosmic-panel-config]
 git = "https://github.com/pop-os/cosmic-panel"
 
+[workspace.dependencies.cosmic-notifications-config]
+git = "https://github.com/pop-os/cosmic-notifications"
+
 [workspace.dependencies.cosmic-randr-shell]
 git = "https://github.com/pop-os/cosmic-randr"
 
--- a/cosmic-settings/Cargo.toml
+++ b/cosmic-settings/Cargo.toml
@@ -25,6 +25,7 @@ cosmic-dbus-networkmanager = { git = "https://github.com/pop-os/dbus-settings-bi
 nm-secret-agent-manager = { git = "https://github.com/pop-os/dbus-settings-bindings", optional = true }
 cosmic-idle-config.workspace = true
 cosmic-panel-config = { workspace = true, optional = true }
+cosmic-notifications-config.workspace = true
 cosmic-protocols = { git = "https://github.com/pop-os/cosmic-protocols", optional = true }
 cosmic-randr-shell.workspace = true
 cosmic-randr = { workspace = true, optional = true }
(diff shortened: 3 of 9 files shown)


maria-komarova:
The screenshot of the new Settings page would be helpful for the UX review.

joshuamegnauth54:
cosmic_settings_01

Screenshot from my (almost) stock test VM.

maria-komarova:
A lot of suggested settings seem too granular to include in the Settings application. 
The plan for the positions of the notifications was to tie it to the placement of the Notifications center applet. It's not currently behaving like this but this is still a plan. 
The only setting out of the settings here that we'd be ready to expose in the UI would be the time the notifications shows on screen with the slider. But having an entire Notifications page for a single setting does not seem like a good idea.

joshuamegnauth54:
Would it be better if I move "Maximum notifications" and "Maximum notifications per app" to an advanced drawer but keep the timeout settings on the main Notifications page?

Also, I agree that only having those three settings would be a bit too sparse for a settings page, but presumably more options will be added later as COSMIC gains more features. :grin:  I'm porting an option from Android/iOS in pop-os/cosmic-notifications#141 which would be a good inclusion to the page if it's merged. A lot of users want an easy way to edit the timeouts (see the attached issues) without having to manually edit the config files.

lucasrossell:
It would be great if there was a way to adjust which applications have permissions to send notifications through this menu. That is the most important use case for me personally.

joshuamegnauth54:
@lucasrossell  Does this fit your use case?: pop-os/cosmic-notifications#141 Or do you mean something more granular?

COSMIC Notifications itself needs support for features like muting notifications or bypassing DND so I can't add those to the settings page yet. I'll try to implement what I can.

maria-komarova:
> It would be great if there was a way to adjust which applications have permissions to send notifications through this menu. That is the most important use case for me personally.

Yes, this is the main purpose of most of the Notifications settings pages in other systems and what I would picture being in Notifications page in COSMIC. I am not sure we can easily implement it right now though which is why I haven't mentioned it. But this is certainly something long-term to look into. 

> Would it be better if I move "Maximum notifications" and "Maximum notifications per app" to an advanced drawer but keep the timeout settings on the main Notifications page?

I don't think we'd want to have it even in the advanced drawer. We're really trying to walk the line between having enough essential settings but not too many. The only cases where we go the advanced route is when there is a lot of need for those. I just don't see the maximum notifications filling much of a need at this point.

joshuamegnauth54:
@maria-komarova I can remove the settings for anchor, max notifications, and max app notifications but leave the timeout settings since users are asking for that. I can change timeout to a slider since that's nicer than what I have now. :laughing: 

I exposed max notifications because the notifications program already supports it. :grin: In other words, I didn't implement that myself. I'll disable it, but I'm positive users will ask for something similar in the future. Three is fairly low default and I already changed it on my personal system. :laughing: 

(I'll comment on the per app notifications thingy on my other PR since it's more relevant there.)

vertexodessa:
@maria-komarova since there is no other way to configure notification duration, I would consider this PR important

tormeh:
Would it be possible to have the timeout for notifications be infinity? I.e. that they have to be manually dismissed to disappear from your screen? Mac OS calls this the persistent alert style. I am a forgetful person, so when a notification is minimized it might just as well be gone...

joshuamegnauth54:
@tormeh I checked COSMIC Notification's source code to see if that's already implemented or at least doable. It's _not_ implemented yet but seems easy to implement. I think the only source code changes would be to add a config item for persistent notifications then to ignore the timeout if it's `true`.

I can try to implement it soon.

@vertexodessa Sorry for the hassle. :sweat_smile: Unfortunately, most of my Notification patches are currently blocked. I'm willing to fix them up if there's still interest in getting them merged though.

tormeh:
I'm unsure if it's cleaner or messier to implement infinity as just the maximum of whatever numeric type the timeout is stored as. If it's longer than a month I doubt anyone would notice the difference, and you get to reuse the same code path.

wash2:
> > It would be great if there was a way to adjust which applications have permissions to send notifications through this menu. That is the most important use case for me personally.
> 
> Yes, this is the main purpose of most of the Notifications settings pages in other systems and what I would picture being in Notifications page in COSMIC. I am not sure we can easily implement it right now though which is why I haven't mentioned it. But this is certainly something long-term to look into.

We could filter displayed notifications based on an allow-list or deny-list using the "desktop-entry" hint. Not all apps would set the value of this hint, but it could still be done with a custom filter for the provided app name too.

joshuamegnauth54:
I implemented persistent notifications here: pop-os/cosmic-notifications#157

For now, you have to manually enable it by writing `true` to the `persistence` settings file.

PR #1894, opened 2026-03-07
fix(network): use from_utf8_lossy for saved SSID decoding

modelmiser:

## Summary

Five callsites in the NM subscription decode saved SSIDs with `String::from_utf8().ok()`, while `current_networks.rs:52` and `available_wifi.rs:47` already use `String::from_utf8_lossy()` for access point SSIDs. This patch aligns all 5 callsites to `from_utf8_lossy`, making SSID decoding consistent within the crate.

The mismatch could cause silent failures for networks with non-UTF-8 SSIDs: the AP list would show them (with replacement characters via `from_utf8_lossy`), but saved-connection lookups would skip them (via `from_utf8().ok()` returning `None`), making forget, connect-to-known, and password retrieval silently fail. Non-UTF-8 SSIDs are rare in practice, but the inconsistency is worth fixing regardless.

## Test plan

- [ ] Verify known networks list still populates correctly
- [ ] Verify forget-network works for a saved WiFi network
- [ ] Verify connect-to-known works (password retrieval path)
- [ ] If a non-UTF-8 SSID is available: verify it appears in both the access point list and the known networks list after saving

### DIFF
--- a/subscriptions/network-manager/src/lib.rs
+++ b/subscriptions/network-manager/src/lib.rs
@@ -452,7 +452,7 @@ async fn start_listening(
                         if s.wifi
                             .clone()
                             .and_then(|w| w.ssid)
-                            .and_then(|ssid| String::from_utf8(ssid).ok())
+                            .map(|ssid| String::from_utf8_lossy(&ssid).into_owned())
                             .is_some_and(|s| s == ssid.as_ref())
                         {
                             _ = c.delete().await;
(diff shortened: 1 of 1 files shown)


modelmiser:
Note: there's one more instance of the same `String::from_utf8(ssid).ok()` pattern at `cosmic-settings/src/pages/networking/wifi.rs:1101`, outside the NM subscription crate. It builds the saved SSID-to-UUID map and would similarly skip non-UTF-8 SSIDs. Worth a follow-up if this PR is merged.

jackpot51:
If I wanted Claude to make this change I would direct it myself

PR #1868, opened 2026-02-24
updates after iced-rebase

wash2:
- [x] I have disclosed use of any AI generated code in my commit messages.
  - If you are using an LLM, and do not fully understand the changes it is making to the code base, do not create a PR.
  - In our experience, AI generated code often results in overly complex code that lacks enough context for a proper fix or feature inclusion. This results in considerably longer code reviews. Due to this, AI authored or partially authored PRs may be closed without comment.
- [x] I understand these changes in full and will be able to respond to review comments.
- [x] My change is accurately described in the commit message.
- [x] My contribution is tested and working as described.
- [x] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

PR #1907, opened 2026-03-16
feat(sound): add output test button with playback fallbacks

### LINKED ISSUE #1901: No TEST button on Cosmic Settings -> Sound-> Output->Output device
No TEST button on Cosmic Settings -> Sound-> Output->Output device
This was available in Gnome Settings. No TEST button under Device Profiles

manascb1344:
---

## Summary
Adds a **Test** button to the Sound Output settings page to verify audio output devices are working. This feature was available in GNOME Settings but missing in COSMIC Settings (issue #1901).

## Before
- Users had no built-in way to test if their selected output device was working
- Had to rely on external applications or media playback to verify audio output
- No feedback mechanism to confirm the correct device was selected

## After
- Added a **Test** button next to the Output device dropdown
- Button is enabled only when an active output sink is available
- Clicking the button plays a test sound through the currently selected output device
- Multiple fallback playback methods ensure compatibility across different system configurations

## Technical Details
**UI Changes:**
- Added `TestOutput` message variant to the Sound page message enum
- Test button appears inline with the Output device selector
- Button is disabled when no active sink is available

**Backend Implementation:**
- `Model::test_output()` spawns async playback task
- Fallback chain for maximum compatibility:
  1. `canberra-gtk-play -i audio-test-signal` (GNOME-compatible sound theme)
  2. `paplay` with multiple test sound paths (`.oga`, `.ogg`, `.wav` formats)
  3. `pw-play` with same test sound paths (PipeWire-native)
- 5-second timeout per command attempt
- Comprehensive tracing for debugging playback failures

**i18n:**
- Added `sound-output.test = Test` localization key (English)
- Other locales will be handled via Weblate

## Visual Comparison

**Before**
image

**After**
image
---

Fixes #1901

### DIFF
--- a/cosmic-settings/src/pages/sound/mod.rs
+++ b/cosmic-settings/src/pages/sound/mod.rs
@@ -34,6 +34,7 @@ pub enum Message {
     SetSinkVolume(u32),
     /// Request to change the input volume.
     SetSourceVolume(u32),
+    TestOutput,
     /// Messages handled by the sound module in cosmic-settings-subscriptions
     Subscription(subscription::Message),
     /// Surface Action
@@ -212,6 +213,10 @@ impl Page {
                     .map(|message| Message::Subscription(message).into());
             }
 
+            Message::TestOutput => {
+                self.model.test_output();
+            }
+
             Message::ToggleOverAmplificationSink(enabled) => {
                 self.amplification_sink = enabled;
 
@@ -334,6 +339,7 @@ fn output() -> Section<crate::pages::Message> {
     crate::slab!(descriptions {
         volume = fl!("sound-output", "volume");
         device = fl!("sound-output", "device");
+        test = fl!("sound-output", "test");
         _level = fl!("sound-output", "level");
         balance = fl!("sound-output", "balance");
         left = fl!("sound-output", "left");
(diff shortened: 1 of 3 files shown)


manascb1344:
The commit https://github.com/pop-os/cosmic-settings/commit/415e1d88329fb6b2cd3dbe1481404374f2000c26 was merged to master without proper formatting, so the CI failure exists

jacobgkau:
@pop-os/ux The contributor provided an "after" screenshot. Would we want this button, and if so, is that where we'd want it?

@manascb1344 It looks like the branch has a conflict, likely because we needed to prioritize https://github.com/pop-os/cosmic-settings/pull/1868. Could you rebase/resolve that?

maria-komarova:
What happens with the button when someone clicks it? Does it get disabled while the sound is playing? 
What happens if no sound is heard? How would people be able to say that the sound should have played but didn't? If we are using Test button here directly we need some kind of indication of what is going on. Maybe the button label should change to a speaker icon while the state of the button changes to disabled. 
Apart from that how does it look when the window is narrow width? Responsiveness is a concern considering the dropdown label could be very long as shown on the screenshot. 

The option to test the output seems potentially useful although I don't see this functionality on all the systems I looked at.

manascb1344:
@maria-komarova, thanks for the review! I'll add a "Testing..." state to the button and fix the layout for narrow windows. Should the Test button go below the dropdown or stay on the same side?

PR #1920, opened 2026-03-20
fix(bluetooth): call connect() after pair() for unpaired devices

hojjatabdollahi:
Calling pair() works for keyboards since they do a whole process when pairing and connect later. For headphones we should connect right away, otherwise the user has to press the connect button twice. Calling pair and connect works for both.

___


hojjatabdollahi:
I noticed there is no Bluetooth section in the QA list, may I suggest adding a section that "forget -> discover -> connect" a bluetooth device to the list?

mmstick (dismissed):

PR #1937, opened 2026-03-27
Quick setting, add autoconnect toggle to VPN connections

Citingmarc:
I've been loving CosimcDE, great work!

I added a couple of incompatible wireguard connections as VPNs through the settings app is that they both would auto connect on login or reboot. Being that both were incompatible I had to disable one of them every restart. I couldn't immediately find the auto connect setting, had to look it up. So I figured since I had a bit of rust experience that it would be fun to try and make it more accessible. Now, not sure if you'll care for this feature or may prefer to keep it on the settings, so feel free to close this if it's not a good fit. But I've found it useful and very nice to work on this codebase :+1: 

image

  - Used AI to help identify patterns in the codebase and assist with writing/reviewing the affected lines. I understand the code and have experience with Rust myself :)
- [x] I have read the [[Developer Certificate of Origin](https://developercertificate.org/)](https://developercertificate.org/) and certify my contribution under its conditions.

### DIFF
--- a/cosmic-settings/src/pages/networking/vpn/mod.rs
+++ b/cosmic-settings/src/pages/networking/vpn/mod.rs
@@ -82,6 +82,8 @@ pub enum Message {
     UsernameUpdate(String),
     /// Display more options for an access point
     ViewMore(Option<ConnectionId>),
+    /// Toggle autoconnect for a VPN connection
+    ToggleAutoconnect(ConnectionId, String, bool),
     /// Create a new wireguard connection
     WireGuardConfig,
     /// Update the text input for the wireguard device name
@@ -132,7 +134,7 @@ impl From<Message> for crate::pages::Message {
 #[derive(Clone, Debug)]
 pub enum ConnectionSettings {
     Vpn(VpnConnectionSettings),
-    Wireguard { id: String },
+    Wireguard { id: String, autoconnect: bool },
 }
 
 #[derive(Clone, Debug, Default)]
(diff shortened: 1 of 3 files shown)


jklgrasso:
I like the idea of making finding the setting simpler. But I noticed a couple of things. 

@pop-os/ux 

### Built from source; testing with two ProtonVPN connections using seperate interfaces (proton0 & proton1):

More than one VPN can have ```Connect Automatically``` toggled on. When both VPNs are on the same interface, only one will connect. The one that connects is likely based on connection priority, or potentially based on whichever one had ```Connect Automatically``` toggled on first in cosmic-settings. 

### Built from source; Testing with two ProtonVPN connections using the same interface (proton0):

When both VPNs are on different interfaces, and both connections have ```Connect Automatically``` enabled, both VPNs connect automatically. This could cause the system to connect to the unwanted VPN, resulting in the original issue. Of course, the user doesn't have to enable both VPNs to connect automatically. But it could happen, potentially creating confusion.

Citingmarc:
Yeah it can still lead to confusing situations. My two cents is that I was happy that they defaulted to auto connect (although it was unexpected). But when adding multiple ones, they both connected and it was very obvious what the problem was, just not how to solve it.

I personally lean towards keeping the drop down as minimal as possible.  Advanced users they can dig into the settings or use nmcli directly if they want to adjust priorities and interfaces. Just that for some people especially light users they may want to control which VPN auto connects if any at all. Especially if they only have one VPN just this single toggle is enough to avoid digging further into any other settings. Some VPNs may significantly slow down the internet, and the default behavior of auto connecting would not be ideal then.

maria-komarova:
It seems all right to me to add this. The menu requires changes though since using toggle is not quite appropriate in this particular menu. 
menu
We should use the list item with checkmark (indent when de-selected). It also looks like the item uses a different text color on the screenshot. Please, make sure it is the same and that all the spacing is correct.

PR #1974, opened 2026-04-21
Change naming of super-key-action option 'disable' to 'none'.

SammyMcFly:
This change was suggested/requested by @maria-komarova in the following comment: https://github.com/pop-os/cosmic-epoch/discussions/3087#discussioncomment-15655231 .

Not all languages have been changed accordingly, yet, as I am not able to verify the correctness for all languages. On that note, the wording of the setting "Super key action" (en) is not perfectly consistent: in some languages, for example Spanish (es), the setting is only called "Super key" (translated into English), which the "disabled" option fits better to, than "none" does. This should probably be made consistent.


jacobgkau (approved):
Verbiage change appears as expected, and all options for this setting still work.

SammyMcFly:
> Just want to confirm, the discussion you linked was about the new Desktop -> Workspaces "Action on typing" setting, which already says "None". The change you're making here would be for the Desktop -> Window management "Super key action" setting, which currently uses "Disable".

Yes, exactly. The relevant quote in the discussion is the second part of the following: "I know Super key action was probably used as an example but we should change it there as well."

jacobgkau:
Yeah, I missed that sentence when I glanced at the page, but deleted my comment after I saw what you were referring to.

SammyMcFly:
> On that note, the wording of the setting "Super key action" (en) is not perfectly consistent: in some languages, for example Spanish (es), the setting is only called "Super key" (translated into English), which the "disabled" option fits better to, than "none" does. This should probably be made consistent.

If desired, I am happy to review the languages that I am acquainted with in regards to wording.

jacobgkau:
> If desired, I am happy to review the languages that I am acquainted with in regards to wording.

That would be appreciated, but all non-English strings need to be edited in [Weblate](https://hosted.weblate.org/projects/pop-os/) rather than directly in the repo.

jacobgkau (dismissed):
Still looks good.

SammyMcFly:
I did changes for the translation of "Super key action" for Spanish, Italian and Russian, as their wording was not aligned with the English, as mentioned before.

When changing the selection label from "Disable" to "None", I would argue to also change the corresponding subfield of "super-key" from ".disable" to ".none" at the same time.

So, as soon as the subfield has changed and you have therefore confirmed that you commit to this naming, I will review the translation of "None", as well.

mmstick (approved):

i18n/ar/cosmic_settings.ftl
```diff
@@ -352,7 +352,7 @@ super-key = إجراءات مفتاح Super
     .launcher = فتح مشغل التطبيقات
     .workspaces = فتح مساحات العمل
     .applications = فتح التطبيقات
-    .disable = تعطيل
+    .none = تعطيل
```
mmstick: Only i18n/en is permitted to be modified. Other files are maintained by weblate.

mmstick (changes requested):

i18n/ar/cosmic_settings.ftl
SammyMcFly: I removed the changes of files in subfolders of "i18n" except "en" from the commit.

mmstick (approved):
Needs extra caution when merging to ensure that Weblate is sync immediately before and immediately after merging since it removes a key.

jacobgkau (approved):

mmstick (approved):

PR #1972, opened 2026-04-21
feat(sound): rebase with varlink sound settings API

mmstick:
- Update to use the varlink sound settings API from the settings daemon.
- Drops the cosmic-pipewire crate that the settings daemon now manages
- Drops the pipewire and pulse subscriptions that will be deprecated by this
- Drops the libpipewire build dependency

Depends on https://github.com/pop-os/cosmic-settings-daemon/pull/135 and https://github.com/pop-os/cosmic-applets/pull/1352

---

- [x] I have disclosed use of any AI generated code in my commit messages.
  - If you are using an LLM, and do not fully understand the changes it is making to the code base, do not create a PR.
  - In our experience, AI generated code often results in overly complex code that lacks enough context for a proper fix or feature inclusion. This results in considerably longer code reviews. Due to this, AI authored or partially authored PRs may be closed without comment.
- [x] I understand these changes in full and will be able to respond to review comments.
- [x] My change is accurately described in the commit message.
- [x] My contribution is tested and working as described.
- [x] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

PR #1979, opened 2026-04-23
fix(sound): match ActiveRoute events to the correct node

tmshlvck:
Fix https://github.com/pop-os/cosmic-applets/issues/1388

When a device has multiple sink/source nodes (e.g. a GPU with multiple HDMI outputs on a single ALSA card), an ActiveRoute event for the currently active device caused the handler to pick the first node from the device_ids IntMap whose device matched. IntMap iteration order is hash-based, so the first node was effectively random among nodes sharing the same device.

Use card_profile_device to disambiguate: only match a node when its card.profile.device is listed in route.devices. This mirrors the matching already done in update_device_route_name.

A LLM was used to analyze the code paths and draft the patch. The diagnosis and fix were verified end-to-end against a real multi-HDMI setup (Radeon HDMI/DP card with three sink nodes) by reproducing the stuck-volume symptom and confirming it no longer occurs after the patch is applied.

### DIFF
--- a/subscriptions/sound/src/lib.rs
+++ b/subscriptions/sound/src/lib.rs
@@ -520,6 +520,15 @@ impl Model {
                 if active_device == Some(id) {
                     for (node_id, &device) in &self.device_ids {
                         if device == id && node_ids.contains(&node_id) {
+                            // Match the route to the correct node via card_profile_device,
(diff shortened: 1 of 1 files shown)


mmstick:
Conflicts with #1972

PR #2014, opened 2026-05-26
feat(power): add controls for power button behavior

### LINKED ISSUE #1664: Missing power button and lid close behavior controls in Power settings
The Power settings page in COSMIC does not currently expose controls for
configuring hardware power actions such as:

- Power button behavior
- Sleep button behavior
- Laptop lid close behavior

These options appear in earlier COSMIC Power panel designs (issue #10),
but are not available in the current Settings UI.

Without these controls, users cannot:
- Prevent accidental shutdowns
- Customize behavior when using external monitors
- Configure different actions for laptop use cases

This issue is specifically about the **missing settings UI**, not underlying
power management behavior.

Reference design / context:
https://github.com/pop-os/cosmic-settings/issues/10

gsboylan:
Adds a section to the power submenu allowing the user to select from the
following power button behaviors:
- Confirm Shutdown (default, existing behavior) - cosmic-osd shutdown
- Confirm Log Out - cosmic-osd logout
- Confirm Restart - cosmic-osd restart
- Shutdown - systemctl poweroff
- Log Out - loginctl terminate-user $USER
- Restart - systemctl reboot
- Suspend - systemctl suspend
- Hibernate - systemctl hibernate

The default option is 'Confirm Shutdown', unchanged from prior to this
commit.

If the value at ~/.config/cosmic/com.system76.CosmicSettings.Shortcuts/v1/system_actions
has been customized to some command outside of this list, that value
will not be overwritten unless the user selects an item in the list
again.

Closes #1664

- [X] I understand these changes in full and will be able to respond to review comments.
- [X] My change is accurately described in the commit message.
- [X] My contribution is tested and working as described.
- [X] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.


cosmic-settings/src/pages/power/backend/mod.rs
```diff
@@ -97,6 +101,92 @@ pub fn get_power_profiles() -> Vec<PowerProfile> {
...
+                None => Some(PowerButtonBehavior::ConfirmShutdown),
+                Some(cmd) if cmd == "cosmic-osd shutdown" => Some(Self::ConfirmShutdown),
+                Some(cmd) if cmd == "cosmic-osd log-out" => Some(Self::ConfirmLogOut),
+                Some(cmd) if cmd == "cosmic-osd restart" => Some(Self::ConfirmRestart),
+                Some(cmd) if cmd == "systemctl poweroff" => Some(Self::Shutdown),
+                Some(cmd) if cmd == "loginctl terminate-user $USER" => Some(Self::LogOut),
+                Some(cmd) if cmd == "systemctl reboot" => Some(Self::Restart),
+                Some(cmd) if cmd == "systemctl suspend" => Some(Self::Suspend),
+                Some(cmd) if cmd == "systemctl hibernate" => Some(Self::Hibernate),
+                // Only return None if the user has customized the command
```
mmstick, line 130: The power button behavior is managed by logind on systemd systems so it should obtain the value from this.  In `/etc/systemd/logind.conf`, there is a `HandlePowerKey` option which defaults to `poweroff`. But you should get the value of this key directly from the DBus service with https://docs.rs/logind-zbus/latest/logind_zbus/manager/struct.ManagerProxy.html#method.handle_power_key.

mmstick (changes requested):

cosmic-settings/src/pages/power/backend/mod.rs
gsboylan: When I was first looking into this, it seemed that cosmic-session inhibits logind from receiving the button press event so it can handle it internally. For my understanding, are you proposing that I retain the logind inhibitor in cosmic-session, but respect/update the logind configuration anyway? 

If the goal is to respect logind.conf, I'm not entirely certain why we'd still want to take the inhibitor lock, and vice-versa if the goal is to handle the button event internally for compat with non-systemd systems then this would leave us with a behavior gap to reconcile. Alternately, maybe there needs to be a specific config option to not take the inhibitor and let the system handle the button press however it's configured.

Happy to make the tweak either way, just want to make sure I understand the desired functionality here. My initial approach was just to edit the existing cosmic shortcut configuration to keep things tidy.

gsboylan: Just so we're on the same page, here's the source for the systemd/logind inhibitor that I'm referring to: https://github.com/pop-os/cosmic-session/blob/master/src/main.rs#L194-L199

gruemelmonster: Any progress on this? Would love to be able to select the powerbutton behaviour (and also the lidswitch:-))

Not that I have anything to say on this, but being able to set it via systemd would probably be the most flexible. But to my understanding of the current code it is indeed inhibited by cosmic-session otherwise I would have easily be able to change it on Arch/CachyOS to my liking.

Issue #843, opened 2024-12-26
Memory leak in `cosmic-settings`


andre4ik3:
After exiting from COSMIC, there were still 3 instances of `cosmic-settings` left running (from me opening & closing it from within COSMIC), then while I was in KDE Plasma, the `cosmic-settings` instances were leaking memory in the background, and two were killed by the kernel OOM killer: [Kernel OOM logs](https://github.com/user-attachments/files/18255626/kernel-oom.txt). The third one I recorded on video, it went up to around 94 GiB and then hovered there. However it would not respond to quit requests, I had to send a SIGKILL for it to close, even after it stopped growing in memory usage.

https://github.com/user-attachments/assets/ff16f73f-9c93-4b4d-a722-101a423a33bb

After I killed it, it sent this in journal:

```
app-cosmic-com.system76.CosmicAppList-344252.scope: Consumed 7min 19.390s CPU time, 94.1G memory peak.
```

OS: latest NixOS unstable, using latest nixos-cosmic flake
Version: `cosmic-settings-1.0.0-alpha.4-unstable-2024-12-17`

bhh32:
Do you have any steps for us to try and replicate the issue? If we can get it to happen again, it may be easier for us to track down where in the code this is stemming from and fix it.

andre4ik3:
I was able to reproduce it just now again on latest 2025.01.02 version. Each cosmic-setting process is using ~25 GiB, for a total of ~100 GiB.

![изображение](https://github.com/user-attachments/assets/5232bef0-591d-4079-9980-46cb2d2902cf)
![изображение](https://github.com/user-attachments/assets/03d37b00-a936-44a4-bae7-02adff937874)

1. Go to COSMIC (my display-manager is SDDM, not sure if it's relevant)
2. Open cosmic-settings, go to "Monitors" section
3. If you want, close and open a few times, it will reopen on "Monitors" section (in my screenshots I closed and reopened 4 times)
4. Exit COSMIC desktop
5. Enter KDE plasma, open system monitor and watch RAM go up (goes up faster the more times you opened/closed cosmic-settings)
6. (Optional) Kill all the cosmic-settings processes quickly before the system OOM's :)

bhh32:
> I was able to reproduce it just now again on latest 2025.01.02 version. Each cosmic-setting process is using ~25 GiB, for a total of ~100 GiB.
> 
> 
> 
[...]

PR #2020, opened 2026-05-28
fix(wallpaper): skip ImageHandle::from_rgba when source and fit unchanged

rchennau:
## Problem

`cosmic-settings desktop` leaks ~9 GB/hour of RSS when running as a persistent background daemon with slideshow mode active. The process is eventually OOM-killed after exhausting swap, which starves the Wayland compositor of CPU time and freezes the display. This is a reproduction of #843.

## Root Cause

`cache_display_image()` calls `ImageHandle::from_rgba()` on every `Message::UpdateState` event (fired once per wallpaper rotation by the inotify watcher on the cosmic-bg state file). `from_rgba` unconditionally calls `Id::unique()`:

```rust
// pop-os/iced core/src/image.rs:216
static NEXT_ID: AtomicU64 = AtomicU64::new(0);
Self(_Id::Unique(NEXT_ID.fetch_add(1, atomic::Ordering::Relaxed)))
```

Every call produces a monotonically-increasing unique ID, making cache hits in the wgpu raster cache (`wgpu/src/image/raster.rs` — `FxHashMap<image::Id, Memory>`) structurally impossible.

Eviction only happens inside `trim()` (`raster.rs:78`), which is only called from `draw()` (`wgpu/src/lib.rs:162`), which is only called from `present()`. When the Wayland compositor suppresses `RedrawRequested` between rotations (DPMS off, occlusion, throttling), `present()` never fires, `trim()` never runs, and the `FxHashMap` grows unbounded — one `Memory` entry per rotation interval.

This is the same `trim()`/`present()` gap described in iced-rs/iced#2659, triggered here by the slideshow rotation path.

## Fix

Add `cached_display_key: Option<DefaultKey>` and `cached_display_fit: Option<usize>` to `Page`. At the start of `cache_display_image()`, resolve the effective `DefaultKey` for the current selection and short-circuit if it matches the previously rendered values:

```rust
if self.cached_display_handle.is_some()
    && resolved_key == self.cached_display_key
    && Some(self.selected_fit) == self.cached_display_fit
{
    return;  // source unchanged — existing handle is valid
}
```

`ImageHandle::from_rgba` is now called only when the wallpaper or fit mode actually changes, not on every state event. All three existing `cached_display_handle = None` sites are updated to clear the new fields so state stays coherent.

## Tested

- Pop!_OS 24.04, cosmic-settings v1.0.12, NVIDIA RTX 5060 Ti + Intel iGPU (hybrid display path)
- Built from source, deployed, and monitored for 20 minutes
- RSS stable at ~243 MB (previously grew at ~9.3 GB/hour)
- Slideshow rotation continues to update the display preview correctly

## Note on complementary fix

This prevents new unique handles from accumulating. A complementary fix in `pop-os/iced` — decoupling `trim()` from `present()` so stale entries are evicted even when the compositor suppresses redraws — would provide defence-in-depth (see iced-rs/iced#2659).

### DIFF
--- a/cosmic-settings/src/pages/desktop/wallpaper/mod.rs
+++ b/cosmic-settings/src/pages/desktop/wallpaper/mod.rs
@@ -155,6 +155,13 @@ pub struct Page {
     /// Cache for storing the image used by the display preview.
     cached_display_handle: Option<ImageHandle>,
 
+    /// The DefaultKey and fit mode of the last successfully rendered display image.
+    /// Used to skip redundant ImageHandle::from_rgba calls: each call produces a unique
+    /// ID via AtomicU64::fetch_add (pop-os/iced core/src/image.rs:216), so calling it
+    /// when the source is unchanged leaks entries into the wgpu raster cache indefinitely.
+    cached_display_key: Option<DefaultKey>,
+    cached_display_fit: Option<usize>,
+
     /// Model for the category dropdown, which has categories and recent folders.
     categories: dropdown::multi::Model<String, Category>,
 
@@ -227,6 +234,8 @@ impl page::Page<crate::pages::Message> for Page {
     fn on_leave(&mut self) -> Task<crate::pages::Message> {
         // Reclaim memory
         self.cached_display_handle = None;
+        self.cached_display_key = None;
+        self.cached_display_fit = None;
         self.selection = Context::default();
         self.outputs = SingleSelectModel::default();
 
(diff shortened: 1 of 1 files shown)


leviport:
The PR template is mandatory

mmstick:
The change isn't valid. Newly-created image handles should have unique IDs. The display preview's handle is dropped when it is updated with a new handle so it's not accumulating here. Not able to replicate memory use increase with the slideshow setting.

PR #2040, opened 2026-06-17
fix(default_apps): set text editor associations for application/x-zerosize

mmstick:
When setting the default text editor, set for both so that empty files without an extension will open with the selected text editor.

Related to https://github.com/pop-os/cosmic-edit/pull/580

---

### DIFF
--- a/cosmic-settings/src/pages/applications/default_apps.rs
+++ b/cosmic-settings/src/pages/applications/default_apps.rs
@@ -231,7 +231,10 @@ impl Page {
                             "x-scheme-handler/https",
                         ],
                     ),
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
Before https://github.com/pop-os/cosmic-edit/pull/580, empty files would open up in e.g. Gedit instead of COSMIC Text once Gedit was installed, even if the default had always been set to COSMIC Text. After https://github.com/pop-os/cosmic-edit/pull/580, empty files would open up in COSMIC Text regardless of the default text editor setting.

With this PR, empty files open up in the application set as the default text editor.

The full regression testing checklist is not needed because only the Default Applications page is being modified.

wash2 (approved):

PR #2073, opened 2026-07-11
feat: Grid workspace layout option with columns/rows settings

AldenDana:
Companion to pop-os/cosmic-comp#2555 (Grid workspace layout, pop-os/cosmic-workspaces-epoch#51).

Adds a **Grid** choice to Desktop → Workspaces → Workspaces orientation, with **Grid columns** / **Grid rows** spin buttons shown while Grid is selected, writing `workspace_grid_columns`/`workspace_grid_rows`. The compositor hot-reloads the change live.

The touchpad gestures description treats Grid like Vertical.

**App workspace assignments.** Also adds a new section to the Workspaces page for pop-os/cosmic-comp#2556's per-app workspace rules (`workspace_assignments`): lists existing rules, lets you remove one, and lets you add a new one (app ID + target workspace number). Previously the only way to manage these was the window context menu or hand-editing the config file; this gives a bulk-management view alongside it.

Depends on the `cosmic-comp-config` changes from pop-os/cosmic-comp#2555 and pop-os/cosmic-comp#2556; to build/test before those merge, add a `[patch]` for `cosmic-comp-config` pointing at those branches.

### DIFF
--- a/cosmic-settings/src/pages/desktop/workspaces.rs
+++ b/cosmic-settings/src/pages/desktop/workspaces.rs
@@ -4,9 +4,11 @@
 // TODO make settings work
 
 use cosmic::cosmic_config::{self, ConfigGet, ConfigSet};
-use cosmic::widget::{self, settings};
+use cosmic::iced::{Alignment, Length};
+use cosmic::widget::{self, icon, settings, text};
 use cosmic::{Apply, Element, surface};
 use cosmic_comp_config::workspace::{Action, WorkspaceConfig, WorkspaceLayout, WorkspaceMode};
+use cosmic_comp_config::WorkspaceAssignment;
 use cosmic_settings_page::{self as page, Section, section};
 use slotmap::SlotMap;
 use tracing::error;
@@ -22,6 +24,11 @@ pub enum Message {
     SetShowName(bool),
     SetShowNumber(bool),
     Surface(surface::Action),
+    SetAssignmentWorkspace(usize, u32),
+    RemoveAssignment(usize),
+    NewAssignmentAppId(String),
+    NewAssignmentWorkspace(u32),
+    AddAssignment,
 }
 
 pub struct Page {
@@ -32,6 +39,9 @@ pub struct Page {
     action_on_typing_active: Option<usize>,
     show_workspace_name: bool,
     show_workspace_number: bool,
+    workspace_assignments: Vec<WorkspaceAssignment>,
+    new_assignment_app_id: String,
+    new_assignment_workspace: u32,
 }
 
 impl Default for Page {
@@ -61,6 +71,15 @@ impl Default for Page {
 
             false
         });
+        let workspace_assignments: Vec<WorkspaceAssignment> = comp_config
+            .get("workspace_assignments")
+            .unwrap_or_else(|err| {
+                if err.is_err() {
+                    error!(?err, "Failed to read config 'workspace_assignments'");
+                }
+
+                Vec::new()
+            });
         Self {
             config,
             comp_config,
@@ -73,6 +92,9 @@ impl Default for Page {
             action_on_typing_active,
             show_workspace_name,
             show_workspace_number,
+            workspace_assignments,
+            new_assignment_app_id: String::new(),
+            new_assignment_workspace: 1,
         }
     }
 }
@@ -87,6 +109,7 @@ impl page::Page<crate::pages::Message> for Page {
             sections.insert(multi_behavior()),
             sections.insert(workspace_orientation()),
             sections.insert(workspace_navigation()),
+            sections.insert(workspace_assignments()),
         ])
     }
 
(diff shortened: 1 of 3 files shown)


AldenDana:
**AI disclosure**

This implementation was developed with assistance from Claude (Anthropic's AI coding assistant), directed and reviewed by me throughout — I specified the requirements, reviewed the generated diffs, and tested every feature live on my own machine. I'm continuing to work with Claude to address review feedback, so turnaround should be quick either way.

  - (Disclosing here in the PR rather than retroactively rewriting already-pushed commit messages.)

mmstick:
Is there a corresponding issue for this which has been approved by @pop-os/ux for implementation?

Also, your commit descriptions are violating the AI policy. If any code within a commit was generated, it must be clearly documented in the commit descriptions. You will need to rebase this with proper citations.

AldenDana:
Fair on both counts.

On the issue: no, there isn't a `@pop-os/ux`-approved issue backing this — I built it as a working prototype (running as my daily driver) to demonstrate the feature and its UX concretely, hoping it'd be useful input to that discussion rather than assuming it was pre-approved. Happy to loop in `@pop-os/ux` here or on pop-os/cosmic-workspaces-epoch#51 if that's the right next step, and to hold off on further review until that's sorted.

On the commits: rebased all four branches (this one, cosmic-comp#2555, cosmic-comp#2556, cosmic-workspaces-epoch#318) with an explicit AI-assisted disclosure appended to every commit message, and force-pushed. No code changes, message-only.

funnylookinhat:
@mmstick Could we have a discussion on https://github.com/pop-os/cosmic-workspaces-epoch/issues/51 about adding this?  It's literally the only feature preventing myself (and several others) from being able to move to COSMIC.

Or - what's the process for discussing this?  I'm happy to jump into a meeting or a chat on Mattermost.

WatchMkr:
@funnylookinhat we decided to [move workspaces into cosmic-comp](https://github.com/pop-os/cosmic-comp/issues/2722) for unrelated reasons but it also blocks additions at the moment.

I think this feature would require a different mode for the current applet. Using the current workspace picker in the overview with a grid can consume the display leaving little space for spread windows - especially if there are a lot of them. I imagine an applet pop-up like from any other applet that shows the grid in a smaller preview above the window spread.

* Four finger gestures and keyboard shortcuts (app movement and navigation) would need to follow the grid layout.
* Arranging workspaces in the grid popup
* Moving workspaces from one display to any other
* Moving apps form the window spread to different workspaces

funnylookinhat:
I think you've nailed the use cases!  Looking forward to the rewrite into `cosmic-comp` - maybe we can revisit it after that work is complete! :+1:

AldenDana:
@WatchMkr @funnylookinhat following up on the requirements above — worth noting this already exists and is running daily on my machine, in case it saves rework once the `cosmic-comp` move happens:

- **4-finger gestures + keyboard nav following the grid layout** — this is solved in [cosmic-comp#2555](https://github.com/pop-os/cosmic-comp/pull/2555): a `SwipeAction::GridMove(Direction)` swipe gesture plus grid-aware `Focus`/`Move` keyboard actions, so Left/Right/Up/Down actually traverse the 2D grid instead of just linear next/prev. This was the part I expected to be the hard part too, but it's been stable for me since July.

The applet-popup idea (small grid preview above the window spread instead of consuming the whole overview) is a great one, though — that UI doesn't exist yet in what I've built. I'll take a look at prototyping it. Thanks for pushing this forward.

AldenDana:
@WatchMkr @funnylookinhat — following up on my note above about prototyping the applet popup: it's built, and I've been running it as my daily driver.

Branches: `AldenDana/cosmic-workspaces-epoch` → `workspace-switcher-popup`, plus the panel button in `AldenDana/cosmic-applets` → `grid-workspace-popup`.

Against your list:

- **Applet popup instead of the fullscreen overview** — a small centred panel showing the live workspace thumbnail grid. It's capped at 80% of the output and the thumbnails scale down as the grid grows, so a wide grid shrinks the tiles rather than eating the display. Clicking outside dismisses it. The existing fullscreen overview is untouched.
- **Moving apps to other workspaces** — done, and slightly beyond the brief: each window gets a grab handle positioned from its real geometry, so a window can be dragged out of *any* workspace tile onto another, not only from the active workspace's spread.
- **Gestures / keyboard nav following the grid** — unchanged, still cosmic-comp#2555.
- **Moving workspaces between displays** — worth flagging that this already works today in the fullscreen overview, and needs nothing new: dragging a workspace onto one belonging to another output sends `MoveWorkspaceBefore`/`After`, and `Shell::move_workspace` already handles `new_output != old_output` by moving it to the other output's group. That's existing upstream behaviour, not something I added.

Not done, to be clear:

- **Arranging workspaces inside the popup.** Two attempts failed. A whole-tile drag source swallows the per-window drags, because `dnd_source` arms on its own bounds without checking whether a child already captured the press — so a nested source always loses. A smaller dedicated grip in the tile footer didn't work out either. Reordering still works in the fullscreen overview.

One thing that may be worth knowing for the `cosmic-comp` move: with `OutputBound` workspaces every output has its own active workspace, and that is easy to get wrong. My first cut of the workspace-switch indicator tracked a single global "active workspace" and misbehaved on a two-monitor setup until every part of it — last-active tracking, the label, and the dismiss timer — was keyed per output.

Fully aware this lands mid-rewrite and isn't a merge request; posting it as input in case the UX or the edge cases are useful.

PR #2098, opened 2026-07-21
fix(sound): source volume changes applied to sink

### LINKED ISSUE #2097: [Bug] Changing Input Volume in Sound Settings Also Changes Output Volume
**Issue/Description**
Changing the **Input** slider in the **Sound** page of COSMIC Settings also changes the **Output** slider to the same value.

Additionally, there is an inconsistency between COSMIC Settings and the Sound applet:

* Adjusting the Input slider from the Sound applet correctly updates the corresponding slider in Settings.
* However, adjusting the Input slider from Settings doesn't affect the Sound Applet: Input slider, instead it changes the Output slider in instead, indicating that the settings page is not correctly updating the input volume.

**Version**
```
Fedora 44 COSMIC: 1.3.0^git20260714.99cf076
```

**Steps to Reproduce**
1. Open Settings.
2. Navigate to Sound.
3. Adjust the Input slider.
4. Observe the Output slider.
5. Compare the behavior with adjusting the Input from the Sound applet.

**Expected Behavior**
* Changing the Input Volume slider in Settings should only affect the input volume.
* The Input should remain synchronized between the Sound applet and Settings app, regardless of which interface is used to make the change.
* Adjusting the input volume should never modify the output volume.

**Other Notes:**

https://github.com/user-attachments/assets/c604015d-bf35-43ce-81ff-6e2ba19093ae

mmstick:
Closes #2097 

---

### DIFF
--- a/cosmic-settings/src/pages/sound/mod.rs
+++ b/cosmic-settings/src/pages/sound/mod.rs
@@ -265,7 +265,7 @@ impl Page {
                         tokio::time::sleep(Duration::from_millis(128)).await;
                         _ = client
                             .conn
-                            .set_sink_volume(volume.load(atomic::Ordering::Relaxed))
+                            .set_source_volume(volume.load(atomic::Ordering::Relaxed))
                             .await;
                         Message::ReattachClient(Arc::new(client)).into()
                     });


wash2 (approved):

jacobgkau (approved):
Looks like this was caused by https://github.com/pop-os/cosmic-settings/pull/2063. Not sure what testing was performed on the sound page for that PR since the notes don't specify any.

The fix works. No other pages are being modified, so the full checklist for the rest of the pages isn't needed.

PR #2100, opened 2026-07-21
feat: add option to toggle shake cursor to find it

hojjatabdollahi:
Depends on:
- [x] https://github.com/pop-os/cosmic-comp/pull/2610

The string for the toggle is from KDE.
It's on by default.

image

____

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1373,7 +1373,6 @@ dependencies = [
 [[package]]
 name = "cosmic-comp-config"
 version = "1.0.0"
-source = "git+https://github.com/pop-os/cosmic-comp#7ee78d9c1369033aa4f963e22d2864bbcda1fbff"
 dependencies = [
  "cosmic-config",
  "input",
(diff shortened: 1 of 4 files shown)


maria-komarova:
I wonder if "Shake mouse pointer to make it bigger" might be a more descriptive copy.

KodeBarista:
In mac it's called "Shake mouse pointer to locate"

image

hojjatabdollahi:
@maria-komarova How about adding a caption for it?

KDE has it as "shake cursor to find it" and MacOS has it has "shake cursor to locate it" so that precendent makes it more recongnizable by the users. But I also understand that your suggestion is more accurate. If you don't like to add the caption (since it's the only option with the caption) I can change the text to what you suggested.


image

leviport (dismissed):
The new option appears and is working as expected. Since this only touches the accessibility page, I only did regression testing there, where I found no issues.

hojjatabdollahi:
@leviport 
I still need to wait for the cosmic-comp part to merge first to update the cosmic-comp-config in cargo.toml. So, after force pushing that we need another review.

But if you can double check with @maria-komarova and make sure the final wording is OK, that would be great.

mmstick (approved):

leviport (approved):
Still looking good after a quick re-test.

PR #2154, opened 2026-08-19
Add back navigation for multiple Bluetooth adapters

### LINKED ISSUE #3931: Bluethooth Settings with  several BLE recivers no backbutton
Hi,

I wanted to let you know that a back button is missing when you have two or more Bluetooth receivers. If you click on one, there is no direct way to go back—you have to click on the Bluetooth menu icon again.

I have attached screenshots below for reference.

Image

Image

Thanks for the great work. 

Tim

alvinwin:
## Summary

- add the standard COSMIC sub-page back control after selecting one of multiple Bluetooth adapters
- return directly to the adapter chooser by clearing the selected adapter
- recompute the Bluetooth status heading when entering or leaving an adapter
- add regression coverage for chooser, selected-adapter, and return states

## Root cause

The multi-adapter chooser stores the selected adapter as in-page state. Bluetooth is registered as a top-level settings page, so it did not receive the automatic back header used by separately registered sub-pages even though `SelectAdapter(None)` already provided the correct state transition.

Fixes pop-os/pop#3931.

## Validation

- `cargo fmt --all --check`
- `cargo clippy --all-features`
- `cargo test --all-features` (42 passed)
- manually launched `cosmic-settings bluetooth` with two adapters, selected an adapter, and used the new back control to return to the chooser

### Manual result

Bluetooth adapter detail with the new back control

The existing mouse-button-4 behavior is outside this patch; this change addresses the missing visible navigation control reported in the issue.

### DIFF
--- a/cosmic-settings/src/pages/bluetooth/mod.rs
+++ b/cosmic-settings/src/pages/bluetooth/mod.rs
@@ -12,13 +12,15 @@ use futures::channel::oneshot;
 use futures::{SinkExt, StreamExt};
 use slotmap::SlotMap;
 use std::collections::{HashMap, HashSet};
-use std::sync::Arc;
+use std::sync::{Arc, LazyLock};
 use std::time::Duration;
 use zbus::zvariant::OwnedObjectPath;
 
 #[cfg(test)]
 use crate::service_manager::MockServiceManager;
 use crate::service_manager::ServiceManagerHandle;
+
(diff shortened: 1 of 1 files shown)


mmstick (approved):

jklgrasso (approved):
Since the Bluetooth page was the only change, I didn't do the full checklist.

#### Bluetooth

- [x] Bluetooth device connects.
- [x] Paired device can be disconnected & reconnected.
- [x] Paired device can be forgotten & connecting again works.

This change works as expected :)

PR #2191, opened 2026-09-08
Add toggle for keyboard layout change OSD

### LINKED ISSUE #3853: Add an option to disable the keyboard layout change OSD
Since the recent update, COSMIC displays an OSD notification every time the keyboard layout is changed, for example "English (US)".

I switch between two keyboard layouts very frequently while typing. Because of this, the notification appears constantly and is very distracting. It can also cover the text or UI elements I am currently working with.

I don't want to disable the entire OSD, because I still want volume and brightness notifications.

Could you please add a setting to disable only the keyboard layout change OSD?

For example:

☐ Show keyboard layout change OSD

This should be an optional setting rather than a mandatory notification.

System:

Pop!_OS 24.04 LTS
COSMIC Desktop
Wayland
cosmic-osd: 0.1.0~1787855233~24.04~aee6ee9

The behavior was introduced by commit:
aee6ee9 — "Add OSD for layboard layout change"

Screenshot attached.

Image

6insanes:
Closes pop-os/cosmic-epoch#3853.

Companion to pop-os/cosmic-osd#226, which adds a
`com.system76.CosmicOsd` / `show_keyboard_layout_osd` cosmic-config
key (default `true`) that `cosmic-osd` now checks before showing its
keyboard-layout-change indicator.

This adds a toggler for that key in the Keyboard settings page, in
the Input Sources section next to the layout list. Reads/writes the
config directly with the same inline `Config::new`/`get`/`set`
pattern already used for `amplification_sink`/`amplification_source`
in the Sound page (no shared struct with `cosmic-osd`, just the same
string key name, so there's no cross-crate dependency between the two
PRs). English localization only; other languages come through
Weblate.

This change was implemented by Claude Code (Anthropic), directed,
reviewed, and tested by the author (isolated KMS test session running
both patched binaries together, plus manual confirmation in the
author's real session).

- [x] I have read the Developer Certificate of Origin and certify my contribution under its conditions.


mmstick:
Requires approval from UX. We have to decide if the feature is necessary.

justF0T0N:
Would realy like to have option to disable that popup

PR #2193, opened 2026-09-09
Add libpop-support-panel to dependencies

thomas-zimmerman:
Support would really, really like to have `pop-support` installed by default. `libpop-support-panel` is where `pop-support` is packaged.


debian/control
```diff
@@ -31,6 +31,7 @@ Depends:
     cosmic-settings-daemon,
     gettext,
     iso-codes,
+    libpop-support-panel,
```
mmstick, line 34: I think `pop-support` would be better since it doesn't have that library anymore.

leviport:
Do we want this installed on non-System76 hardware?

debian/control
thomas-zimmerman: I just added the package that has the `pop-support` command: 
```
dpkg -S /usr/bin/pop-support
libpop-support-panel: /usr/bin/pop-support
```
Is the packaging going to change for 26.04?

mmstick: It's an alias that I might remove

thomas-zimmerman:
I don't think we did in 22.04 and having the logs even on non-sys76 hardware makes sense for community support too.

mmstick:
We may add support/repair functions to the utility that our apps can integrate with. Currently thinking about migrating the repair functions from pop-upgrade to this if we integrate it better.

PR #2088, opened 2026-07-19
cursor: Add UI for idle-hide timeout

herder:
Adds the cosmic-settings UI for the `cursor_hide_timeout` compositor option introduced in [pop-os/cosmic-comp#2393](https://github.com/pop-os/cosmic-comp/pull/2393) (implements [pop-os/cosmic-comp#2231](https://github.com/pop-os/cosmic-comp/issues/2231)).

Partially addresses [pop-os/cosmic-settings#603](https://github.com/pop-os/cosmic-settings/issues/603) — implements the "hide after inactivity" half; the separately-requested "hide cursor when typing" option is out of scope here.

**Location:** Settings → Input → Mouse

- A **"Hide cursor when inactive"** toggle, mapping to `Some`/`None` on the `cursor_hide_timeout` key.
- A **spin button** for the timeout in seconds (min 1, max 300, step 1), matching the `active_hint`/`gaps` spinners on the appearance page.

The seconds value is retained in page state while the toggle is off, so toggling off then on restores the previous value rather than resetting it. A seed of 5 s is used the first time it is enabled (the compositor default remains disabled/`None`). Persisted through the input page's existing `com.system76.CosmicComp` config handle; cosmic-comp live-reloads the key.

- [x] I have disclosed use of any AI generated code in my commit messages.
- [x] I understand these changes in full and will be able to respond to review comments.
- [x] My change is accurately described in the commit message.
- [x] My contribution is tested and working as described.
- [x] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

Issue #603, opened 2024-09-17
Feature Request: Configurable Mouse Cursor Hiding


davxy:
It would be great to have an option in the settings to hide the mouse cursor after a configurable time period of inactivity. Additionally, there could be an option to immediately hide the cursor when the user starts typing. This feature is quite popular among users of tiling window managers (such as Sway and Hyprland), where the mouse is often secondary to keyboard navigation.

Use Case:

For users who rely heavily on keyboard input and use tiling, the mouse cursor can often become a distraction when it stays visible on the screen after periods of inactivity. By providing a setting to automatically hide the cursor after a configurable timeout, or immediately when typing begins, the user experience can be significantly improved, especially for those using keyboard-driven workflows.

Proposed Settings:

1. Hide cursor after inactivity:

Type: Boolean (Enable/Disable)
Timeout: Configurable duration (e.g., 5 seconds, 10 seconds, etc.)
Default: Disabled

2. Hide cursor when typing:

Type: Boolean (Enable/Disable)
Default: Disabled

Thank you for considering this request!

awilkins:
+1 for this : 

Especially considering that moving focus with the keys also brings the mouse pointer along with it and parks it centre screen, right in the middle of e.g. your multi-pane `vim` window.

el4ra:
+1
Heavy user of tiling with COSMIC on NixOS. Would love the ability to hide the cursor (or at least keep it in place) especially when jumping from screen to screen using Super+Arrows. Right now the cursor will go to the center of the screen and stay there.

virackson:
Is there any possibility of this feature being implemented? The inactive mouse cursor tends to be a bit distracting, and I often find myself needing to move it manually to some corner. Thank you so much for considering!

cureforboredom:
+1

I can deal with the jank and bugs, that honestly doesn't annoy me and I am loving cosmic.

I just need my mouse cursor to get out of the way when I'm or otherwise not using it. For whatever reason, VSCode doesn't hide the cursor, which is incredibly frustrating, but on X I could use xbanish. Please consider this feature.

mmj2023:
Yeah, all other wayland compositors that I have tried have the option to auto hide the cursor. Hopefully, this gets added soon. But, is there a manual way like changing the config or running a command to set the option?

wezm:
It looks like this was implemented in https://github.com/pop-os/cosmic-comp/pull/2393 but there is no UI to control it yet. You can manually set the config value though:

```
[...]

PR #2196, opened 2026-09-12
input: Add settings for contextual cursor hiding

herder:
Adds the UI for the contextual cursor hiding introduced in [pop-os/cosmic-comp#2849](https://github.com/pop-os/cosmic-comp/pull/2849), which reworks [#2088](https://github.com/pop-os/cosmic-settings/pull/2088) along the lines @maria-komarova asked for there. Closes the "hide when typing" half of [#603](https://github.com/pop-os/cosmic-settings/issues/603), and the "hide after inactivity" half.

**Location:** Settings → Input → Mouse, as a new "Hiding the cursor" section.

| control | config field |
| --- | --- |
| Hide while typing | `while_typing` |
| Hide after touch input | `after_touch` |
| Hide in fullscreen + delay | `fullscreen_idle_timeout` |
| Hide when inactive + delay | `idle_timeout` |

**Every trigger gets its own switch, deliberately.** The concern raised on #2088 was that users could unintentionally lose the cursor. Two of these triggers are on by default in the compositor, so a cursor that vanishes with no visible way to stop it is exactly that disorientation — making each one findable and switchable is what answers it, rather than exposing fewer controls.

**The touch row only appears when udev reports a touchscreen**, matching the existing `system_has_touchpad` gate that hides the touchpad sub-page. There is nothing to configure on a machine that cannot generate touch input.

**Each delay's seconds are retained in page state while its switch is off**, so switching back on restores the previous value rather than resetting. The compositor stores "off" as `None` and therefore remembers no previous value itself; seeds are 5 s for inactivity and 3 s for fullscreen.

## Blocked on the compositor PR

This uses `cosmic_comp_config::CursorHideConfig`, which does not exist upstream yet — `cosmic-comp-config` is sourced from `git+https://github.com/pop-os/cosmic-comp` and pinned in `Cargo.lock`. **It will not build until [pop-os/cosmic-comp#2849](https://github.com/pop-os/cosmic-comp/pull/2849) merges and that rev is bumped**, which is why this is a draft. `Cargo.lock` is deliberately untouched here so the bump is a single reviewable commit once the dependency is available.

Defaults live in the compositor, not here, and are an open question for @pop-os/ux on that PR — this UI reflects whatever it decides without changes.

## Testing

Three unit tests on the state mapping: a disabled timeout is still seeded, seconds survive a switch off and on, and enabling one timeout does not disturb the other.

---

- [ ] My contribution is tested and working as described.


herder:
Screenshot:
image

PR #2198, opened 2026-09-12
show display identifiers

FreddyFunk:
The display identifiers OSDs where originally added in https://github.com/pop-os/cosmic-osd/pull/151 and as part of implementing https://github.com/pop-os/cosmic-settings/issues/885 . Unfortunately the related previous implementation in cosmic-settings https://github.com/pop-os/cosmic-settings/pull/1554 caused regressions and was reverted in https://github.com/pop-os/cosmic-settings/pull/1605 .

This new implementation instead passes cosmic-settings existing display ordering directly to cosmic-osd, avoiding the output-management feedback loop.

depends on https://github.com/pop-os/cosmic-osd/pull/229

### DIFF
--- a/cosmic-settings/src/app.rs
+++ b/cosmic-settings/src/app.rs
@@ -274,6 +274,12 @@ impl cosmic::Application for SettingsApp {
     }
 
     fn on_app_exit(&mut self) -> Option<Self::Message> {
+        #[cfg(feature = "page-display")]
+        if self.pages.page_id::<display::Page>() == Some(self.active_page)
+            && let Some(display) = self.pages.page_mut::<display::Page>()
+        {
+            _ = display.deactivate_display_identifiers();
+        }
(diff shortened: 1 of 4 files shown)


cosmic-settings/src/app.rs
```diff
@@ -274,6 +274,12 @@ impl cosmic::Application for SettingsApp {
     }
 
     fn on_app_exit(&mut self) -> Option<Self::Message> {
+        #[cfg(feature = "page-display")]
+        if self.pages.page_id::<display::Page>() == Some(self.active_page)
+            && let Some(display) = self.pages.page_mut::<display::Page>()
+        {
+            _ = display.deactivate_display_identifiers();
```
mmstick, line 281: Perhaps this should be applied in a drop impl so that an application panic could still call this.

